PR #1, opened 2021-01-26
Transform interface and example Power transformation

glennmoy:
PoC for the package interface (modelled off Impute.jl) and a prototype transform.


rofinn:
I feel like maybe this package should be called `Transforms.jl` so that we can just have a `Transform` type?

glennmoy:
> I feel like maybe this package should be called `Transforms.jl` so that we can just have a `Transform` type?

Yeah perhaps that's better

src/transformers.jl
```diff
@@ -0,0 +1,66 @@
...
+function transform end
+
+"""
+    transform!(A::AbstractArray{T}, ::Transformation; dims=:, kwargs...) where T <: Real
+
+Applies the transformation to each element of `A`.
+Optionally specify the `dims` to apply the transformation along certain dimensions.
+"""
+function transform!(
+    A::AbstractArray{T}, t::Transformation; dims=:, kwargs...
```
willtebbutt, line 35: I'm picking a line at random here, but my comment is more general.

The present design requires that `kwargs` are propagated through the entire pipeline. This both creates clutter, and means that special-casing is required for things like `Colon()`.

We had a similar problem in [`KernelFunctions.jl`](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl) where the initial design assumed that all inputs to functions were matrices, and you have to propagate an `obsdim` argument everywhere to say whether matrices we to be treated as "N x D" or "D x N" (D=dims, N=number of observations). The solution was to wrap the data itself in a thing that declared which way it was treated. This remove all kwargs and cleaned up the internals considerably.

A similar thing could be done here, specifically by
1. `Transformation`s do whatever transformation they encode to the entire object by default,
2. create `Transformation`s that apply to specific bits of data by composition. For example, introduce an `ApplyTo` `Transformation` `a` that eats another `Transformation` `t` and a "location specifier" -- calling `transform!` on `a` and data `x` would first pull out the bit of data in `x` that `a`'s "location specifier" refers to, and passes it off to `t`. `t` can then apply its transformation to the entirety of the data that it sees.

This would allow you to construct quite complicated transformations by composition e.g. specify that a transformation should only act on certain elements of a certain column by composing two `ApplyTo` transformations or something.

Is this a design choice that you would consider?

willtebbutt: Ohhhh, sorry, I had misunderstood how the present design works. I see what the `transform!` vs `_transform!` are doing now.

IIUC all stuff to do with "which bit of the data am I applying this to" lives in `transform!`, leaving `_transform!` free to operate on all of the data that is passed to it.

I'm still a little confused by the need to propagate kwargs everywhere other than the top layer though. Clearly you need to specify the dimensions to which a thing applies, so `transform!` needs them, but it's not clear to me that you need them in `_transform!` -- I would naively have imagined that the `Transformation` itself would contain all of the information needed to to apply itself to data. Is there a good example of where this isn't the case?

src/transformers.jl
```diff
@@ -0,0 +1,66 @@
...
+
+transform(x, t::Transformation; kwargs...) = transform!(_trycopy(x), t; kwargs...)
+
+"""
+    transform!(table::T, ::Transformation; cols=nothing)::T where T
+
+Applies the transformation to each of the specified columns in the `table`.
+If no `cols` are specified, then the transformation is applied to all columns.
+"""
+function transform!(table::T, t::Transformation; cols=nothing)::T where T
```
bencottier, line 52: I haven't seen `f()::T` syntax before - is that specifying the return type?
Why doesn't the Array transform have that?

test/power.jl
```diff
@@ -0,0 +1,42 @@
+@testset "power" begin
```
bencottier, line 1: We should test non-default `dims` and `cols`.

src/power.jl
```diff
@@ -0,0 +1,12 @@
...
+    Power(exponent) <: Transformation
+
+Raise the data by the given `exponent`.
+"""
+struct Power <: Transformation
+    exponent::Real
+end
+
+function _transform!(x::AbstractArray{T}, P::Power; kwargs...) where T <: Real
+    x[:] = x.^ P.exponent
```
bencottier, line 11: I get `InexactError` if I do
```
x = [1, 2, 3]
p = Power(1.5)
p(x)
```
because it tries to do `Int()` on the elements of `p(x)`.

src/transformers.jl
bencottier: > I would naively have imagined that the Transformation itself would contain all of the information needed to to apply itself to data
Just noting I also expected this.

src/transformers.jl
```diff
@@ -0,0 +1,66 @@
...
+"""
+abstract type Transformation end
+
+# Make Transforms callable types
+(t::Transformation)(x; kwargs...) = transform(x, t; kwargs...)
+
+"""
+    transform!(data::T, transformation::Transformation; kwargs...) -> T
+
+Apply `transformation` to the `data` in-place.
```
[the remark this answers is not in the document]
nicoleepp: Could just clarify this docstring by saying it mutates the input instead of using "in-place"

src/transformers.jl
nicoleepp: It does indeed specify the return type. It's not required but will raise an error if the function returns something not of type T

src/utils.jl
```diff
@@ -0,0 +1,9 @@
+function _trycopy(data)
```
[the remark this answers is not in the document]
nicoleepp: Just an additional comment that according to bluestyle function names this should be `_try_copy`

src/Transform.jl
```diff
@@ -1,5 +1,12 @@
 module Transform
 
-# Write your package code here.
+using Tables
+
+export Transformation, Power
```
rofinn, line 5: I wouldn't mind seeing a functional form of all of our transforms for eager evaluation on the repl. For example, you can do `impute(data, LOCF(...))` or `Impute.locf(data; kwargs...)` which is kinda nice. I could see a `@transform` macro that also generates this functional form along with a kwarg constructor.

rofinn (commented):
Makes sense to me, though I'm obviously biased. I think some more extensive testing and sample transforms would help. We may also need to do some iteration on the `kwargs` and function names at a later time.

src/transformers.jl
willtebbutt: > The way I see it is, Transform -> transform is equivalent to ModelFactory -> Model in that sense.

Oh interesting -- this wasn't obvious to me.

As I understand it, the main reason that we make a distinction between `ModelFactory` and `Model` (and stuff along those lines generally) is because we want to apply the same `ModelFactory` to multiple distinct data sets, the details of which aren't known when the `ModelFactory` is constructed. We tend to build model factories outside of the context in which they'll get used, and combine them with problem-specific data when it becomes available (data set dimensions etc).

Conversely, I wasn't imagining that a `Transform` would generally get used in that way. In particular, unless I'm mistaken, we have the data to which a given `Transform` will be applied at construction time because we'll be building them inside `Forecasters.transform`. Or am I mistaken?

rofinn: So we could do that, but I was thinking that we might want to declaratively specify the transforms in the type. The default Forecaster.transform function would just run a default chain of transforms. This should make it more transparent to end users what operations are being applied, and ideally make it possible to modify that chain without rewriting the entire transform function.

src/transformers.jl
bencottier: > Why doesn't the Array transform have that?

So is it because there is particular risk of mutating the type with tables compared to arrays?

src/transformers.jl
glennmoy: >In Impute.jl most of the methods aren't data specific and the dims keyword is pretty ubiquitous for all methods, so pushing that to the shared fallback methods made sense.

I'll mention that basing the design off of Impute was a deliberate choice, made for a few reasons
* Impute has gone through a few iterations in response to how it's been used - so we know it's been field tested
* We expect Transform and Impute to be interoperable, .e.g. compose chains of Imputors/Transforms

I'm open to tweaking the design but it would ideally be something that benefitted Impute as well.

src/transformers.jl
glennmoy: Yeah, for instance certain operations might return a `SubDataFrame`, `DataFrameRow` or just extract the raw data and return a `Matrix`. We'd want to prevent this from happening.

src/transformers.jl
```diff
@@ -0,0 +1,66 @@
...
+"""
+function transform! end
+
+"""
+    transform(data::T, transformation::Transformation; kwargs...) -> T
+
+Non-in-place version of [`transform!`](@ref), which it delegates to by default.
+Does not need to be extended unless in-place transformation is not possible.
+"""
+function transform end
```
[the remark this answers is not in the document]
glennmoy: Yeah I'm not sure...

On the one had we're kinda stuck with `transform` for the sake of consistency in our downstream dependencies (Forecasters).

But on the other hand, `transform` doesn't make sense in the context of `Imputor`s and `apply` would be better there. 

I wonder is there a way to distinguish how these methods/terms apply in certain contexts that makes intuitive sense?
e.g. you `apply` a Transform/Imputor to a certain data type, but you should overload `transform` when combining these operations in a pipeline?

So that `apply(data, ::Union{Transform, Imputor})` make sense but `apply(::MyModel, data)` doesn't. Conversely, `transform(data, ::Union{Transform, Imputor})` doesn't make (total) sense but `transform(::MyModel, data)` does?

src/transformers.jl
```diff
@@ -0,0 +1,66 @@
...
+    A::AbstractArray{T}, t::Transformation; dims=:, kwargs...
+) where T <: Real
+    dims == Colon() && return _transform!(A, t; kwargs...)
+
+    return map(eachslice(A; dims=dims)) do x
+        return _transform!(x, t; kwargs...)
+    end
+end
+
+transform(x, t::Transformation; kwargs...) = transform!(_trycopy(x), t; kwargs...)
```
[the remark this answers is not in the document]
glennmoy: yup, I only needed it when testing the NamedTuple

test/power.jl
```diff
@@ -0,0 +1,42 @@
...
+        expected = (a = [1, 8, 27], b = [64, 125, 216])
+
+        @test transform(nt, p) == expected
+        @test p(nt) == expected
+
+        _nt = deepcopy(nt)
+        transform!(_nt, p)
+        @test _nt == expected
+    end
+
```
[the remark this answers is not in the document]
glennmoy: 😱

src/Transform.jl
glennmoy: Yeah I agree, Distances does this as well I believe, e.g. `euclidean`, which would be quite convenient.

test/power.jl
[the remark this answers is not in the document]
glennmoy: added some rudimentary tests but I think we'll need some robust test utils like Impute has.

src/transformers.jl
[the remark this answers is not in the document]
bencottier: That distinction of `transform`/`apply` for model/data sounds reasonable to me Glenn.

Is it the case that we've made Transforms callable types here, but this is not so for Imputors? I'm wondering if the `transform`/`apply` distinction matters much externally if `transform` is almost always used via the callable type.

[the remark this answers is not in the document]
glennmoy: Yeah I think it would, because you'd want Impute to extend the Transforms API, which is only possible if they use the same function underneath.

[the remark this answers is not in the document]
rofinn: I'm confused, so what would Impute.jl be extending `transform` or `apply`? NOTE: If the API was changed to `transform(MyType, data)` then I'd need to update and deprecate almost every method in Impute.jl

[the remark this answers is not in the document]
glennmoy: Sorry, I'm also working this out as I type it.

>Impute.jl be extending transform or apply?

Impute would extend `apply`.

> If the API was changed to `transform(MyType, data)` then I'd need to update and deprecate almost every method in Impute.jl

Sorry, I could have been clearer. I'm not suggesting we change the API to this. 
I was just using that example (with `MyModel`) to showcase how we use currently use (and overload) `transform`, which is owned by the Forecasters API.

So if we want `Transform` to also own the `transform` method as well (which I'm guessing it should), then I'm trying to use our current application of it as an example of how we might want to distinguish it from `apply`.

e.g.
`Transform.apply` is a (potentially private?) method that is overloaded for new `Transform`/`Imputor`s.
`Transform.transform` is just a public interface method that is overloaded on types that want to perform feature engineering using `Transform`s/`Imputor`s.

Note: I'm possibly trying too hard to shoe-horn in both `apply` and `transform` into this package so let me know if this doesn't make much sense. Especially when trying to justify `transform` my basis is that:
1. We are using a method with that name already and so it should belong somewhere.  
b) It may as well be owned by `Transforms` because we can then define/enforce its contract at this higher level rather than in a dependent package like Forecaster (which isn't the only potential use-case in the long-term).

Does this make sense?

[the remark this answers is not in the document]
glennmoy: I've renamed `transform`->`apply` and added _new_ methods for `transform` with docstrings that I hope explain what I'm getting at.

src/transformers.jl
```diff
@@ -7,61 +7,79 @@ Abstract supertype for all Transforms.
...
 
-Non-mutating version of [`transform!`](@ref), which it delegates to by default.
-Does not need to be extended unless a mutating Transform is not possible.
+Non-mutating version of [`transform!`](@ref).
 """
 function transform end
 
 """
-    transform!(A::AbstractArray{T}, ::Transform; dims=:, kwargs...) where T <: Real
+    Transforms.apply!(data::T, Transform::Transform; kwargs...) -> T
```
bencottier, line 31:
```suggestion
    Transforms.apply!(data::T, ::Transform; kwargs...) -> T
```
?

src/transformers.jl
```diff
@@ -7,61 +7,79 @@ Abstract supertype for all Transforms.
...
+
+Applies the [`Transform`](@ref) mutating the input `data`.
+Where possible, this should be extended for new data types `T`.
+"""
+function apply! end
+
+"""
+    Transforms.apply(data::T, Transform::Transform; kwargs...) -> T
```
bencottier, line 39:
```suggestion
    Transforms.apply(data::T, ::Transform; kwargs...) -> T
```
?

src/transformers.jl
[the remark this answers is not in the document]
bencottier: > Note: I'm possibly trying too hard to shoe-horn in both apply and transform into this package so let me know if this doesn't make much sense. Especially when trying to justify transform my basis is that:
>
>    We are using a method with that name already and so it should belong somewhere.
>    b) It may as well be owned by Transforms because we can then define/enforce its contract at this higher level rather than in a dependent package like Forecaster (which isn't the only potential use-case in the long-term).

It does feel somewhat awkward that the package is handling both individual data transformations and feature engineering pipelines (though the latter only in abstract). It seems like the new role of `transform` is just a contract and won't get more complicated within this package, right? If that's the case I think the awkwardness is outweighed by the reasons you give.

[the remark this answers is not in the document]
rofinn: Yeah, we may find that we can generalize things such that we don't need both in the future,  but that distinction makes sense to me. Anyways, thanks for clarify, this should be pretty easy to extend in Impute.jl.

test/power.jl
```diff
@@ -0,0 +1,95 @@
...
+        @testset "dims = $d" for d in (Colon(), 1, 2)
+            @test Transforms.apply(A, p; dims=d) == expected
+        end
+
+    end
+
+    @testset "AxisKey" begin
+        A = KeyedArray([1 2 3; 4 5 6], foo=["a", "b"], bar=["x", "y", "z"])
+        expected = AxisArray([1 8 27; 64 125 216], foo=["a", "b"], bar=["x", "y", "z"])
```
bencottier, line 71:
```suggestion
        expected = KeyedArray([1 8 27; 64 125 216], foo=["a", "b"], bar=["x", "y", "z"])
```

src/transformers.jl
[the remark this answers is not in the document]
glennmoy: > It seems like the new role of transform is just a contract and won't get more complicated within this package, right?

yes this is correct - there might be certain interface tests we'll eventually write but for now it just says "Transforms owns the transform method".

test/power.jl
glennmoy: now why didn't that fail...

glennmoy: Maybe it's because `==` on `KeyedArrays`/`AxisArrays` / `AbstractArrays` generally  only compares the values and not the types? anyway fixed it.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,94 @@
...
+  pull_request:
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - "1.5"  # Invenia Prod version
```
glennmoy, line 18: Testing on 1.0 complained about a lot of stuff so I didn't think there would be much gain in testing on it.

I feel these are sufficient?

nicoleepp: The Project.toml should bump the existing julia compat: `julia = "1"` if this package doesn't run on 1.0

nicoleepp (approved):

PR #8, opened 2021-02-02
Add LinearCombination transform

### LINKED ISSUE #5: Linear Combination transform
A transform that computes a linear combination of columns in an array.

MWE
```julia
x = rand(3, 2)

lc = LinearCombination([0.1, 0.9])

Transform.apply(x, lc; cols=[1, 2])
```

A few things:

First.
This is probably just a repackaging of [`Statistics.mean`](https://juliastats.org/StatsBase.jl/stable/means/#Statistics.mean) which is just a weighted sum.
Instead of calling this inside `LinearCombination` might it be better/more flexible to make a Transform that accepts generic functions as transformations? 

Second.
Since this is a many-to-one transform, doesn't seem possible that a mutating form is possible.

nicoleepp:
Closes #5 

Note that I couldn't get a linear combination using [Statistics.mean](https://juliastats.org/StatsBase.jl/stable/means/#Statistics.mean) to work. A lot of operations were returning NaNs. Also, that mean function expects a whole matrix of weights and not just the weights for each column. For these reasons, I opted not to define the CustomTransform instead of doing this explicitly.

From the issue description, I am unclear what the role of `dims` kwarg is or how it should work.

### DIFF
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -1,7 +1,7 @@
 """
     LinearCombination(coefficients) <: Transform
 
-Calculate the linear combination using the column weights passed in.
+Calculate the linear combination using the vector coefficients passed in.
 """
 struct LinearCombination <: Transform
     coefficients::Vector{Real}
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -18,28 +18,42 @@ end
 
 _sum_row(row, col_weights) = sum(map(*, row, col_weights))
 
-function apply(x::AbstractVector, LC::LinearCombination; kwargs...)
+"""
+    apply(x, LC::LinearCombination; cols=Colon())
+
+Applies the [`LinearCombination`](@ref) to each of the specified columns in `x`.
+If no `cols` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
+"""
+function apply(x::AbstractVector, LC::LinearCombination; cols=Colon())
     # Treat each element as it's own column
     # Error if dimensions don't match
-    num_cols = length(x)
+    num_cols = cols isa Colon ? length(x) : length(cols)
     _check_dimensions_match(LC, num_cols)
 
-    return [_sum_row(x, LC.col_weights)]
+    return [_sum_row(x[cols], LC.col_weights)]
 end
 
-function apply(x::AbstractArray, LC::LinearCombination; kwargs...)
+function apply(x::AbstractArray, LC::LinearCombination; cols=Colon())
     # Error if dimensions don't match
-    num_cols = size(x, 2)
+    num_cols = cols isa Colon ? size(x, 2) : length(cols)
     _check_dimensions_match(LC, num_cols)
 
-    return [_sum_row(row, LC.col_weights) for row in eachrow(x)]
+    return [_sum_row(row[cols], LC.col_weights) for row in eachrow(x)]
 end
 
-function apply(x, LC::LinearCombination; kwargs...)
+function apply(x, LC::LinearCombination; cols=nothing)
     # Error if dimensions don't match
-    column_names = Tables.columnnames(x)
-    num_cols = length(column_names)
+    num_cols = cols === nothing ? length(Tables.columnnames(x)) : length(cols)
     _check_dimensions_match(LC, num_cols)
 
-    return [_sum_row(row, LC.col_weights) for row in Tables.rows(x)]
+    # Keep the generic form when not specifying column names
+    # because that is much more performant than selecting each col by name
+    if cols === nothing
+        return [_sum_row(row, LC.col_weights) for row in Tables.rows(x)]
+    else
+        return [
+            _sum_row([row[cname] for cname in cols], LC.col_weights)
+            for row in Tables.rows(x)
+        ]
+    end
 end
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -1,0 +1,45 @@
+"""
+    LinearCombination <: Transform
+
+Calculate the linear combination using the column weights passed in.
+"""
+struct LinearCombination <: Transform
+    col_weights::Vector{Real}
+end
+
+function _check_dimensions_match(LC::LinearCombination, num_cols)
+    num_weights = length(LC.col_weights)
+    if num_cols != num_weights
+        throw(DimensionMismatch(
+            "Number of cols ($num_cols) doesn't match number of weights ($num_weights)"
+        ))
+    end
+end
+
+_sum_row(row, col_weights) = sum(map(*, row, col_weights))
+
+function apply(x::AbstractVector, LC::LinearCombination; kwargs...)
+    # Treat each element as it's own column
+    # Error if dimensions don't match
+    num_cols = length(x)
+    _check_dimensions_match(LC, num_cols)
+
+    return [_sum_row(x, LC.col_weights)]
+end
+
+function apply(x::AbstractArray, LC::LinearCombination; kwargs...)
+    # Error if dimensions don't match
+    num_cols = size(x, 2)
+    _check_dimensions_match(LC, num_cols)
+
+    return [_sum_row(row, LC.col_weights) for row in eachrow(x)]
+end
+
+function apply(x, LC::LinearCombination; kwargs...)
+    # Error if dimensions don't match
+    column_names = Tables.columnnames(x)
+    num_cols = length(column_names)
+    _check_dimensions_match(LC, num_cols)
+
+    return [_sum_row(row, LC.col_weights) for row in Tables.rows(x)]
+end
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -35,14 +35,18 @@ function apply(x::AbstractVector, LC::LinearCombination; inds=Colon())
 end
 
 """
-    apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
+    apply(x::AbstractMatrix, LC::LinearCombination; dims=1, inds=Colon())
 
 Applies the [`LinearCombination`](@ref) to each of the specified indices in `x` along the
-dimension specified, which defaults to applying across the columns of x.
+dimension specified, which defaults to applying it row-wise for each column of x."
 
 If no `inds` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
 """
-function apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
+function apply(x::AbstractMatrix, LC::LinearCombination; dims=1, inds=Colon())
+    if dims === Colon()
+        throw(ArgumentError("Colon() dims is not supported, use 1 or 2 instead"))
+    end
+
     # Get the number of vectors in the dimension not specified
     other_dim = dims ==  1 ? 2 : 1
     num_inds = inds isa Colon ? size(x, other_dim) : length(inds)
@@ -53,24 +57,24 @@ function apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
 end
 
 """
-    apply(x::Table, LC::LinearCombination; inds=nothing
+    apply(x::Table, LC::LinearCombination; cols=nothing)
 
-Applies the [`LinearCombination`](@ref) to each of the specified indices (columns) in `x`.
+Applies the [`LinearCombination`](@ref) to each of the specified cols in `x`.
 
-If no `inds` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
+If no `cols` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
 """
-function apply(x, LC::LinearCombination; inds=nothing)
+function apply(x, LC::LinearCombination; cols=nothing)
     # Error if dimensions don't match
-    num_inds = inds === nothing ? length(Tables.columnnames(x)) : length(inds)
-    _check_dimensions_match(LC, num_inds)
+    num_cols = cols === nothing ? length(Tables.columnnames(x)) : length(cols)
+    _check_dimensions_match(LC, num_cols)
 
     # Keep the generic form when not specifying column names
     # because that is much more performant than selecting each col by name
-    if inds === nothing
+    if cols === nothing
         return [_sum_row(row, LC.coefficients) for row in Tables.rows(x)]
     else
         return [
-            _sum_row([row[cname] for cname in inds], LC.coefficients)
+            _sum_row([row[cname] for cname in cols], LC.coefficients)
             for row in Tables.rows(x)
         ]
     end
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -1,58 +1,76 @@
 """
-    LinearCombination <: Transform
+    LinearCombination(coefficients) <: Transform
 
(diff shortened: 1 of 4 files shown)


glennmoy:
>From the issue description, I am unclear what the role of dims kwarg is or how it should work.

I think `dims` would apply when you want the linear combination to only operate on a subset of columns.

e.g. 
```julia
df = DataFrame(
    :time = Day(1):Day(1):Day(5),
    :a = [1, 2, 3, 4, 5],
    :b = [1, 1, 2, 3, 5],
)

lc = LinearCombination([1, 5])

Transforms.apply(df, lc; cols=[:a, :b])
```
similarly for `dims` on arrays. 
Does this make sense?

bencottier:
> similarly for `dims` on arrays.

If it's selecting a subset of columns, then `dims` for arrays seems misleading - like it will apply the linear combination along a different axis. I think it should always be called `cols` for this usage.

test/linear_combination.jl
```diff
@@ -0,0 +1,96 @@
...
+
+        # Does this just test dims is ignored?
+        @testset "dims = $d" for d in (Colon(), 1, 2)
+            @test Transforms.apply(M, lc; dims=d) == expected
+            @test lc(M; dims=d) == expected
+        end
+
+        @testset "dimension mismatch" begin
+           M = [1 1 1; 2 2 2]
+           @test_throws DimensionMismatch Transforms.apply(M, lc)
```
bencottier, line 34:
```suggestion
            M = [1 1 1; 2 2 2]
            @test_throws DimensionMismatch Transforms.apply(M, lc)
```

bencottier (commented):
We should handle `dims`/`cols` as Glenn explained, but otherwise this looks good to me.

nicoleepp:
> > similarly for `dims` on arrays.
> 
> If it's selecting a subset of columns, then `dims` for arrays seems misleading - like it will apply the linear combination along a different axis. I think it should always be called `cols` for this usage.

I agree using `dims` would make me thing it's choose only certain rows, not a subset of the array cols

nicoleepp:
I have implemented `cols` as Ben suggested and left the `dims` tests commented out while we decide what `dims` should mean. I may just not understand what `dims` means for an Array but having both does seem confusing.

bencottier (approved):
Looks good to me. I don't think there is a use for `dims` as long as this transform is only applied on columns.

glennmoy:
>If it's selecting a subset of columns, then dims for arrays seems misleading - like it will apply the linear combination along a different axis. I think it should always be called cols for this usage.

> Looks good to me. I don't think there is a use for `dims` as long as this transform is only applied on columns.

Sorry, that's dead right, in my first response I completely confused `dims` / `cols`.

But I think there is still a place for `dims` because we might not always take the linear combination column-wise.

e.g. consider just taking a `sum` over some matrix - this resembles a `LinearCombination(ones(3))` in this case.
 
```julia

julia> sum(A, dims=2)
3×1 Array{Int64,2}:
  6
 15
 24

julia> sum(A, dims=1)
1×3 Array{Int64,2}:
 12  15  18
```

I think all that needs to change is to change `eachrow` to `eachslice(...; dims=dims)`?

src/linear_combination.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    LinearCombination <: Transform
+
+Calculate the linear combination using the column weights passed in.
+"""
+struct LinearCombination <: Transform
+    col_weights::Vector{Real}
```
glennmoy, line 7: This is perhaps a nitpick but I believe the proper term for this in a [linear combination context](https://en.wikipedia.org/wiki/Linear_combination) is "coefficients".

I believe "weights" come with additional meaning in ML/statistics of being proportional quantities (probabilities, frequencies, etc.) and usually sum to 1

src/linear_combination.jl
```diff
@@ -0,0 +1,59 @@
...
+
+    return [_sum_row(x[cols], LC.col_weights)]
+end
+
+function apply(x::AbstractArray, LC::LinearCombination; cols=Colon())
+    # Error if dimensions don't match
+    num_cols = cols isa Colon ? size(x, 2) : length(cols)
+    _check_dimensions_match(LC, num_cols)
+
+    return [_sum_row(row[cols], LC.col_weights) for row in eachrow(x)]
```
glennmoy, line 41:
```suggestion
    return [_sum_row(row[cols], LC.col_weights) for row in eachslice(x; dims=dims)]
```

here is where `dims` could come into play

src/linear_combination.jl
```diff
@@ -0,0 +1,59 @@
...
+end
+
+function apply(x, LC::LinearCombination; cols=nothing)
```
glennmoy, line 44: `dims`, correctly, wouldn't make sense here

src/linear_combination.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    LinearCombination <: Transform
```
glennmoy, line 2:
```suggestion
    LinearCombination(coefficients) <: Transform
```

or just `coeffs`, whichever you think

src/linear_combination.jl
```diff
@@ -0,0 +1,77 @@
...
+_sum_row(row, coefficients) = sum(map(*, row, coefficients))
+
+"""
+    apply(x::AbstractVector, LC::LinearCombination; inds=Colon())
+
+Applies the [`LinearCombination`](@ref) to each of the specified indices in `x`.
+
+If no `inds` are specified, then the [`LinearCombination`](@ref) is applied to all elements.
+"""
+function apply(x::AbstractVector, LC::LinearCombination; inds=Colon())
```
nicoleepp, line 28: I left `dims` out for AbstractVectors as it doesn't make sense for them

src/linear_combination.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+"""
+    apply(x::Table, LC::LinearCombination; inds=nothing
+
+Applies the [`LinearCombination`](@ref) to each of the specified indices (columns) in `x`.
+
+If no `inds` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
+"""
+function apply(x, LC::LinearCombination; inds=nothing)
```
nicoleepp, line 62: I was unsure whether `cols` should be renamed to `inds` for consistency even though for tables we do actually want the cols.

test/linear_combination.jl
```diff
@@ -0,0 +1,202 @@
...
+    @testset "Matrix" begin
+        M = [1 1; 2 2; 3 5]
+        expected = [0, 0, -2]
+
+        @testset "all inds" begin
+            @test Transforms.apply(M, lc) == expected
+            @test lc(M) == expected
+        end
+
+        # TODO: Colon() is not supported by eachslice(A, dims=:), do we care?
```
nicoleepp, line 43: What about this? I believe the Power tests use Colon() for dims which is why this question came up

src/linear_combination.jl
morris25: Can we not just generalize the `AbstractArray` code to handle a one dimensional array?

src/linear_combination.jl
```diff
@@ -0,0 +1,77 @@
...
+    apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
+
+Applies the [`LinearCombination`](@ref) to each of the specified indices in `x` along the
+dimension specified, which defaults to applying across the columns of x.
+
+If no `inds` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
+"""
+function apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
+    # Get the number of vectors in the dimension not specified
+    other_dim = dims ==  1 ? 2 : 1
```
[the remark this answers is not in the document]
nicoleepp: Assuming it I think, not sure how it would work for multi dimensinal arrays

test/linear_combination.jl
glennmoy: Interesting.... I'm not sure how to interpret a linear combination (with a vector of coefficients) taken across all dimensions...

Do you apply the LC to one dimension first, then the other? and get a scalar as a result?

In the above cases what we're doing is reducing the dimension of the input by 1, so Matrix-> Vector, Vector->Scalar.
To do that you need to specify the dimension across which you're reducing the elements.

So I don't think (for vector coefficients) we can support `Colon()` since it's undefined which dimension you're taking the sum over.

Does this make sense?

src/linear_combination.jl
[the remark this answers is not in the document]
glennmoy: my mind bends thinking about higher dimensional arrays at the best of times, let alone 8pm on a friday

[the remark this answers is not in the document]
morris25: If we only want to define a LinearCombination for matrices, then we could just require `dims` to be a tuple of which row and column we are doing the transformation over for it to work for any multidimensional array.

src/linear_combination.jl
nicoleepp: I was running into problems bc Vectors and Matrices work slightly differently, also a Vector  is not an AbstractMatrix which is what we are restricting the other method to now

src/linear_combination.jl
```diff
@@ -35,14 +35,18 @@ function apply(x::AbstractVector, LC::LinearCombination; inds=Colon())
 end
 
 """
-    apply(x::AbstractArray, LC::LinearCombination; dims=1, inds=Colon())
+    apply(x::AbstractMatrix, LC::LinearCombination; dims=1, inds=Colon())
 
 Applies the [`LinearCombination`](@ref) to each of the specified indices in `x` along the
-dimension specified, which defaults to applying across the columns of x.
+dimension specified, which defaults to applying it row-wise for each column of x."
```
glennmoy, line 41:
```suggestion
dimension specified, which defaults to applying it row-wise for each column of x.
```

test/linear_combination.jl
```diff
@@ -82,14 +88,21 @@
         end
 
         @testset "dims" begin
+            @testset "dims = :" begin
+                d = Colon()
+                @test_throws ArgumentError Transforms.apply(A, lc; dims=d)
+            end
+
             @testset "dims = 1" begin
```
glennmoy, line 96: does this also work if passing in the axis key? e.g. `foo`, `bar`?

test/linear_combination.jl
```diff
@@ -118,14 +131,21 @@
...
```
glennmoy, line 139: same for `foo`, `bar`?

glennmoy (approved):

test/linear_combination.jl
glennmoy: I didn't check it with `Power` so if it doesn't work don't worry about it

nicoleepp: I doubt it

bencottier: How do you do the same with AxisArrays though? If I use `(Colon(), :foo, :bar)` like for KeyedArray in `test/power.jl` I get

```
dims = foo: Error During Test at /Users/bencottier/JuliaEnvs/Transform.jl/test/power.jl:65
  Got exception outside of a @test
  MethodError: no method matching length(::Symbol)
  Closest candidates are:
    length(!Matched::Base.Iterators.Flatten{Tuple{}}) at iterators.jl:1061
    length(!Matched::Base.MethodList) at reflection.jl:872
    length(!Matched::DataStructures.EnumerateAll) at /Users/bencottier/.julia/packages/DataStructures/ixwFs/src/multi_dict.jl:96
    ...
  Stacktrace:
   [1] #eachslice#196 at ./abstractarraymath.jl:496 [inlined]
   [2] apply!(::AxisArray{Int64,2,Array{Int64,2},Tuple{Axis{:foo,Array{String,1}},Axis{:bar,Array{String,1}}}}, ::Power; dims::Symbol, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:57
   [3] apply(::AxisArray{Int64,2,Array{Int64,2},Tuple{Axis{:foo,Array{String,1}},Axis{:bar,Array{String,1}}}}, ::Power; kwargs::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:dims,),Tuple{Symbol}}}) at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:64
   [4] macro expansion at /Users/bencottier/JuliaEnvs/Transform.jl/test/power.jl:66 [inlined]
   [5] macro expansion at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1190 [inlined]
   [6] macro expansion at /Users/bencottier/JuliaEnvs/Transform.jl/test/power.jl:65 [inlined]
   [7] macro expansion at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115 [inlined]
   [8] macro expansion at /Users/bencottier/JuliaEnvs/Transform.jl/test/power.jl:62 [inlined]
   [9] macro expansion at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115 [inlined]
   [10] top-level scope at /Users/bencottier/JuliaEnvs/Transform.jl/test/power.jl:3
   [11] include(::String) at ./client.jl:457
   [12] top-level scope at /Users/bencottier/JuliaEnvs/Transform.jl/test/runtests.jl:12
   [13] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [14] top-level scope at /Users/bencottier/JuliaEnvs/Transform.jl/test/runtests.jl:11
   [15] include(::String) at ./client.jl:457
   [16] top-level scope at none:6
   [17] eval(::Module, ::Any) at ./boot.jl:331
   [18] exec_options(::Base.JLOptions) at ./client.jl:272
   [19] _start() at ./client.jl:506
```

glennmoy: it looks like `eachslice` doesn't work on AxisArrays maybe. Nevermind it was only a passing thought.

glennmoy: yeah it doesn't work without some extra voodoo

```julia
julia> eachslice(A, dims=axisdim(A, Axis{:foo}))
Base.Generator{Base.OneTo{Int64},Base.var"#199#202"{AxisArray{Int64,2,Array{Int64,2},Tuple{Axis{:foo,Array{String,1}},Axis{:bar,Array{String,1}}}},Tuple{},Tuple{Colon}}}(Base.var"#199#202"{AxisArray{Int64,2,Array{Int64,2},Tuple{Axis{:foo,Array{String,1}},Axis{:bar,Array{String,1}}}},Tuple{},Tuple{Colon}}([1 2 3; 4 5 6], (), (Colon(),)), Base.OneTo(2))
```

And it's been on their todo list for 5 years 🙄 
https://github.com/JuliaArrays/AxisArrays.jl/issues/7

nicoleepp: I can make this work if we want to, by checking if it's an AxisArray. I might prefer to do this in a follow up MR just to keep diffs smaller.

nicoleepp: https://github.com/invenia/Transforms.jl/issues/13

PR #10, opened 2021-02-02
Add Periodic transform

### LINKED ISSUE #2: Periodic transform
A transform that applies a periodic function to the data with arguments
* `f` - the periodic function to apply
* `period` - the period of the function `f`
* `phase_shift` - an optional phase shift to be applied to the function argument 

Note that `phase_shift` is not strictly necessary for this function from a feature engineering perspective but may be required for debugging purposes and/or comparison against our current use of this function.

MWE
```julia
x = 1:1000

# period/phase could also just be positional args
periodic = Periodic(cos, period=5, phase_shift=2)

Transforms.apply!(x, periodic)
```

bencottier:
Closes #2 

The `Periodic` transform is implemented both for `Real` and `TimeType` data, with `Real` and `Period` type fields for the transform, respectively.

The method dispatch might not be the best - some of it was done to fix problems with method ambiguity or not dispatching as intended.

I wasn't sure what to do about mutating tables for the `TimeType`-based transform - it returns an array of columns rather than mutating the DataFrame.


src/periodic.jl
```diff
@@ -0,0 +1,34 @@
...
+    Periodic(f::PeriodicFunction, period<:PeriodicInput) -> Periodic
+
+A constructor for [`Periodic`](@ref).
+Returns a `Periodic` transform with zero phase shift.
+"""
+function Periodic(f, period)
+    return Periodic{typeof(period)}(f, period, zero(typeof(period)))
+end
+
+function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
```
bencottier, line 31: Should we throw a custom error message for `period`/`phase_shift` not matching the data eltype? Or is there another way to enforce that?

src/periodic.jl
```diff
@@ -0,0 +1,34 @@
...
+    x[:] = P.f.(2π .* x ./ P.period .+ P.phase_shift)
```
bencottier, line 32: Adding phase shift seems to be the more common convention (e.g. [Wikipedia article](https://en.wikipedia.org/wiki/Sine_wave)), but I think the old FeatureEngineering function effectively subtracts?

src/periodic.jl
nicoleepp: Does it error anyway if they don't match?

src/periodic.jl
nicoleepp: Do they perform the same? If they do what does benchmarking say is more performant?

src/periodic.jl
bencottier: Yeah, e.g. `ERROR: MethodError: no method matching /(::Float64, ::Day)`. That's probably clear enough, but I think handling it similar to `_check_dimensions_match` for LinearCombination could help.

src/periodic.jl
bencottier: I don't see why they would differ in performance, but they will return different values, because it changes the direction that the function is shifted. Whichever direction we choose can be reversed by passing a negative value for `phase_shift`. So my question relates more to what our users expect phase shift to do. 

I think I will change it to subtract, because (a) one can just change the sign of the phase shift argument to match their expectations (b) phase shift usually isn't needed for feature engineering purposes (c) it aligns with our current implementation so there's less confusion/updating usage.

src/periodic.jl
nicoleepp: Yeah, the LinearCombination didn't error if they didn't match and that's why I added the explicit check

nicoleepp (approved):

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+Applies [`Periodic`](@ref) to each of the specified columns in `table`.
+If no `cols` are specified, then [`Periodic`](@ref) is applied to all columns.
+Returns an array containing each transformed column.
+"""
+function apply(table, P::Periodic{T}; cols=nothing) where T <: Period
+    Tables.istable(table) || throw(MethodError(apply, (table, P)))
+
+    columntable = Tables.columns(table)
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [apply(getproperty(columntable, cname), P) for cname in cnames]
```
bencottier, line 54: Rethinking this - should it return a DataFrame or NamedTuple just to give the columns names, even though this method is agnostic to the specific table type? @nicoleepp

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
+const PeriodicFunction = Union{typeof(cos), typeof(sin)}
+const PeriodicParameter = Union{Real,Period}
+
+"""
+    Periodic(f, period, phase_shift) <: Transform
+
+Applies a periodic function `f` with provided `period` and `phase_shift` to the data.
+
+# Fields
+* `f::PeriodicFunction`: the periodic function
```
glennmoy, line 10:
```suggestion
* `f::Union{typeof(cos), typeof(sin)}`: the periodic function
```

else interpolate the type in the docs using `@docs` but this is probably easier

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
+const PeriodicFunction = Union{typeof(cos), typeof(sin)}
```
glennmoy, line 1: [blue style](https://github.com/invenia/BlueStyle#global-variables) recommends capitalising these.

But I'm personally indifferent in this case since it's a type.

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+const PeriodicParameter = Union{Real,Period}
+
+"""
+    Periodic(f, period, phase_shift) <: Transform
+
+Applies a periodic function `f` with provided `period` and `phase_shift` to the data.
+
+# Fields
+* `f::PeriodicFunction`: the periodic function
+* `period<:PeriodicParameter`: the duration it takes for the periodic function to repeat.
```
glennmoy, line 11:
```suggestion
* `period<:PeriodicParameter`: the function period. Must be non-zero.
```

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+    The sign of the value determines the direction of the function. Must be non-zero.
```
glennmoy, line 12:
```suggestion
```

these seem contradictory?

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+* `phase_shift<:PeriodicParameter`: adjusts the phase of the periodic function, measured
+    in the same units as the input. Increasing the value translates the function to the
+    right, toward higher/later input values.
+"""
+struct Periodic{T<:PeriodicParameter} <: Transform
+    f::PeriodicFunction
+    period::T
+    phase_shift::T
+end
```
glennmoy, line 21: perhaps define an inner constructor that checks that `period` is strictly positive?

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+function Periodic(f, period)
+    return Periodic{typeof(period)}(f, period, zero(typeof(period)))
+end
+
+function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
+    x[:] = P.f.(2π .* (x .- P.phase_shift) / P.period)
+    return x
+end
+
+function apply(x::AbstractArray{T}, P::Periodic{U}; kwargs...) where {T <: TimeType, U <: Period}
```
glennmoy, line 38: I guess you overloaded `apply` because this can't be mutating since it's consuming a time-type put returning a real-type?

src/periodic.jl
glennmoy: It should return a table - since an array isn't expected in this instance - so maybe a NamedTuple with the same column names is appropriate?

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+"""
+function apply(table, P::Periodic{T}; cols=nothing) where T <: Period
+    Tables.istable(table) || throw(MethodError(apply, (table, P)))
+
+    columntable = Tables.columns(table)
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [apply(getproperty(columntable, cname), P) for cname in cnames]
+end
+
+function _periodic(
```
glennmoy, line 57: could use a brief docstring for context

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+    f::PeriodicFunction,
+    instant::TimeType,
+    period::Period,
+    phase_shift::Period=Day(0)
+)
+    # Save the direction of `period` while using abs(period)
```
glennmoy, line 63: It doesn't make sense to me for period to have a direction?

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
...
+    elseif phase_shift < zero(typeof(period))
+        while period_begin < instant
+            period_begin += period  # Make sure `period_begin <= instant`
+        end
+    end
+
+    period_end = period_begin + period
+    # `period_end - period_begin` converts to the same units so that `/` is defined
+    normalized = (instant - period_begin) / (period_end - period_begin)
+    return f(2π * normalized * direction)
```
glennmoy, line 84: it's only being used here to potentially flip the sign of the argument...but why...

Doing this only affects sine 
sin(-x) = -sin(x)
cos(-x) = cos(x)

but I don't see how encoding this in the period is at all intuitive / reasonable...especially when a phase shift / multiplication factor can achieve the same thing more transparently.

src/periodic.jl
bencottier: Hmm, yeah I prefer camel case for types so it's easier to tell visually.

src/periodic.jl
bencottier: Maybe you are thinking non-negative? That would be contradictory. I'm saying it can be negative or positive, but not equal to 0.

src/periodic.jl
bencottier: See other comment, but I've allowed `period` to be negative. The reason is that `FeatureEngineering.periodic` allows this. But it's only a convenience (you can negate the data or change `phase_shift` to get the same effect), and I don't think it matters from a feature engineering standpoint, so it could be dropped.

src/periodic.jl
bencottier: Correct. And the `U <: Period` was needed to avoid method ambiguity with the `apply(table...` method. I was surprised at that because I thought `::AbstractArray{T}` would be enough to resolve that ambiguity.

src/periodic.jl
bencottier: Yeah makes sense. I created a broader issue about this #12. It also raises the question of transformed column naming, but I can use the same column names for now.

src/periodic.jl
glennmoy: yeah, non-negative, my mistake

test/periodic.jl
```diff
@@ -0,0 +1,332 @@
...
+            )
+        end
+
+        @testset "phase shift" begin
+            @test _periodic(sin, DateTime(1), Week(1)) == 0
+            @test _periodic(sin, DateTime(1), Week(1), Day(7)) == 0
+            @test _periodic(sin, DateTime(1), Week(1), Day(-7)) == 0
+
+            # Using the phase shift we can make a Thursday the beginning of the week.
+            # Note: that these calculations can change depending on the period type given.
```
glennmoy, line 59: why is this? seems weird to me. `Second(Week(1))` is the same amount of time as `Week(1)`

test/periodic.jl
```diff
@@ -0,0 +1,332 @@
...
+                -_periodic(sin, DateTime(2000, 1, 1, 3), Day(1)),
+            )
+
+            @test ==(
+                _periodic(cos, DateTime(2000, 1, 1, 3), Day(-1)),
+                _periodic(cos, DateTime(2000, 1, 1, 3), Day(1)),
+            )
+        end
+
+        @testset "phase shift" begin
```
glennmoy, line 53: can we add a test where the result is not 0?

test/periodic.jl
```diff
@@ -0,0 +1,332 @@
...
+
+    @testset "Real input" begin
+        @testset "$f" for f in (sin, cos)
+            p = Periodic(f, 5, 2)
+
+            @test p isa Transform
+
+            @testset "Vector" begin
+                x = collect(0.:11.)
+                expected = f.(2π .* (x .- 2) ./ 5)
```
glennmoy, line 79: can you hard-code this? this is essentially the same code in the function being tested. So if there's a bug in that (unlikely, but still) we won't catch it here. Same for any other tests that call `f.(2π .* (x .- x') ./ d)`

test/periodic.jl
```diff
@@ -0,0 +1,332 @@
+@testset "periodic" begin
+    # Constructors
+    @test Periodic(sin, 5) == Periodic(sin, 5, 0)
```
glennmoy, line 3: maybe test trying to construct a `Periodic(sin, 1, Day(1))` and that the fields are correctly assigned?
```julia
p.period == 5
p.phase_shift == 0
```

src/periodic.jl
```diff
@@ -0,0 +1,85 @@
+const PeriodicFunction = Union{typeof(cos), typeof(sin)}
+const PeriodicParameter = Union{Real,Period}
+
+"""
+    Periodic(f, period, phase_shift) <: Transform
```
glennmoy, line 5:
```suggestion
    Periodic{T}(f, period::T, phase_shift::T) <: Transform
```

might look too cluttered

src/periodic.jl
glennmoy: >The reason is that FeatureEngineering.periodic allows this.

yeah and I'm equally confused about that. I don't see the justification for it especially since

>you can negate the data or change phase_shift to get the same effect

>I don't think it matters from a feature engineering standpoint, so it could be dropped.

I'd like to drop it tbh - if we discover a use case later that needs it we can add it back but I can't seem to find where it's used.

test/periodic.jl
bencottier: Ah, I was lazy - the whole `_periodic` test set is copied from the original `periodic`. So I'm confused too. I need to re-evaluate the old assumptions.

src/periodic.jl
glennmoy: although look at the same function in https://github.com/invenia/Transforms.jl/pull/8

https://github.com/invenia/Transforms.jl/pull/8/files#diff-e6cb55b7b445a33fcc97a850d6c3907c75fdd150531bf7b42f04aae186b9b869R44-R54

which sums the data from a collection of cols, how would we name the return in this instance? 

It starts to get confusing/complicated when you consider the types of transformations we have (one-to-one, many-to-one, etc.) and trying to come up with rules that apply to all.

So maybe reconsider my suggestion for a NamedTuple for now. What do you think @nicoleepp ?

bencottier: For the record I'm keeping this as-is.

src/periodic.jl
```diff
@@ -18,6 +17,13 @@ struct Periodic{T<:PeriodicParameter} <: Transform
     f::PeriodicFunction
     period::T
     phase_shift::T
+
+    function Periodic(f, period, phase_shift)
+        if period <= zero(typeof(period))
+            throw(ArgumentError("period must be strictly positive."))
+end
```
glennmoy, line 24: indenting is off here

test/periodic.jl
bencottier: Doing this, for NamedTuple it looks like I need to use `collect(nt) ≈ collect(nt_expected) atol=1e-14` since there is no `isapprox` for NamedTuples.

I might be able to adjust the hard-coded values to match the output, so that we can use `==` (initially I got them from applying sin/cos in the REPL), but I'm not sure if that will be stable.

test/periodic.jl
bencottier: It's because of `floor(dt:DateTime, p::Period)` in `_periodic`. It gives the nearest multiple of `p` in time.

`floor(DateTime(1), Week(1)) == 0001-01-01T00:00:00`

This is kinda weird because it's a Saturday - I thought it would give a Monday. But it's the year 0001 AD so maybe something weird happens to the calendar. It floors to Monday for recent dates e.g. 

```
floor(DateTime(2021), Week(1)) == 2020-12-28T00:00:00
floor(DateTime(2021, 2, 10), Week(1)) == 2021-02-08T00:00:00
```

Anyway, for seconds:

`floor(DateTime(1), Second(Week(1))) == 0000-12-30T00:00:00`

Because December 30 is exactly 52 weeks _in seconds_ from time 0.

bencottier: For now I've briefly explained this in the `_periodic` docstring.

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+* `f`: the periodic function
+* `period`: the function period. Results can change depending on its type, e.g. `Week(1)`
+    starts the function on Monday (with 0 `phase_shift`), while `Second(Week(1))` starts
+    the function on the nearest multiple of 604800 seconds from time 0.
+* `phase_shift`: adjusts the phase of the periodic function, measured in the same units as
+    the input. Increasing the value translates the function to the right, toward
+    higher/later input values.
+"""
+function _periodic(f, instant, period, phase_shift=Day(0))
+    period = abs(period)
```
glennmoy, line 75:
```suggestion
```

we've already established it's strictly positive

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+    columntable = Tables.columns(table)
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [apply(getproperty(columntable, cname), P) for cname in cnames]
+end
+
+"""
+    _periodic(f, instant, period, phase_shift=Day(0))
+
+Computes the value of periodic function `f` at the given instant in time.
+
```
glennmoy, line 64:
```suggestion
!!! note
    The result will change depending on the type of `period` given, even if the same 
    amount of time is described. Example: `Week(1)` vs `Second(Week(1))`; the former
    returns the previous Monday, while the latter returns the nearest multiple of 604800 
    seconds since time 0.
```

I would put this note here. It's easy to miss when buried in the argument descriptions.

Watch out for the line limit here - I can't tell when in suggestion-mode.

test/periodic.jl
glennmoy: ah ok - makes then.

test/periodic.jl
glennmoy: >I might be able to adjust the hard-coded values to match the output,

I wouldn't bother tbh. rounding errors are to be expected on `sin` / `cos` so `isapprox` and checking the type is fine

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    Periodic{T}(f, period::T, phase_shift::T) <: Transform
+
+Applies a periodic function `f` with provided `period` and `phase_shift` to the data.
+
+# Fields
+* `f::Union{typeof(cos), typeof(sin)}`: the periodic function
+* `period<:Union{Real, Period}`: the function period. Must be strictly positive.
```
glennmoy, line 8:
```suggestion
* `period::Union{Real, Period}`: the function period. Must be strictly positive.
```

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+* `phase_shift<:Union{Real, Period}`: adjusts the phase of the periodic function, measured
```
glennmoy, line 9:
```suggestion
* `phase_shift::Union{Real, Period}`: adjusts the phase of the periodic function, measured
```

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+    in the same units as the input. Increasing the value translates the function to the
+    right, toward higher/later input values.
+"""
+struct Periodic{T} <: Transform where T <: Union{Real, Period}
+    f::Union{typeof(cos), typeof(sin)}
+    period::T
+    phase_shift::T
+
+    function Periodic(f, period, phase_shift)
```
glennmoy, line 18:
```suggestion
    function Periodic(f, period::T, phase_shift::T) where T
```

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+        if period <= zero(typeof(period))
+            throw(ArgumentError("period must be strictly positive."))
+        end
+        return new{typeof(period)}(f, period, phase_shift)
```
glennmoy, line 22:
```suggestion
        return new{T}(f, period, phase_shift)
```

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+                    ZonedDateTime(2015, 11, 1, tz"America/Winnipeg") + Hour(h),
+                    Day(1),
+                )
+                expected = f(2π / 25 * h)
+                @test result ≈ expected atol=1e-14
+            end
+        end
+
+        @testset "phase shift" begin
+            @test _periodic(sin, DateTime(1), Week(1)) == 0
```
glennmoy, line 39: as we've seen below `DateTime(1)` can be weird (start of week is a Saturday)

So might be better to provide modern `DateTime`s in the range we expect to use

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+            @test _periodic(sin, DateTime(1), Week(1), Day(7)) == 0
+            @test _periodic(sin, DateTime(1), Week(1), Day(-7)) == 0
+
+            @test _periodic(sin, DateTime(2), Week(1)) ≈ 0.7818314824680298 atol=1e-14
```
glennmoy, line 43: I'm fine with the test but this level of precision is probably not necessary and could well be affected by rounding errors

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+                0.30901699437494745,
+               -0.8090169943749473,
+               -0.8090169943749475
+            ]
+        )
+
+        @testset "$f" for f in (sin, cos)
+            p = Periodic(f, 5, 2)
+
+            @test p isa Transform
```
glennmoy, line 98: I'd put this under `@testset "Constructors"`

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+
+            @testset "Vector" begin
+                x = collect(0.:5.)
+                expected = expected_dict[f]
```
glennmoy, line 102: you can pull this out of the inner loops and specify it once for all of them

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+                p = Periodic(sin, Day(5))
+                @test p.period == Day(5)
+                @test p.phase_shift == Day(0)
+            end
+        end
+
+        @testset "$f" for f in (sin, cos)
+            p = Periodic(f, Day(5), Day(2))
+
+            @test p isa Transform
```
glennmoy, line 244: same

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+            p = Periodic(sin, Day(5), Day(2))
+            x = 0.:11.
+            @test_throws MethodError Transforms.apply(x, p)
+        end
+
+        @testset "TimeType data, Real transform" begin
+            p = Periodic(sin, 5, 2)
+            x = ZonedDateTime(2020, 1, 1, tz"EST") .+ (Day(0):Day(1):Day(5))
+            @test_throws MethodError Transforms.apply(x, p)
+        end
```
glennmoy, line 367: also test trying to construct `Periodic(sin, 1, Day(3))` and vice versa

src/periodic.jl
bencottier: Should we put that in the `Periodic` docstring so it's external, adding that it's relevant to `TimeType` input?

test/periodic.jl
```diff
@@ -0,0 +1,369 @@
...
+    @testset "_periodic" begin
+        @testset "$f" for f in (sin, cos)
+            # A typical 24-hour day
+            for h in 0:24
+                result = _periodic(
+                    f,
+                    ZonedDateTime(2015, 6, 1, tz"America/Winnipeg") + Hour(h),
+                    Day(1),
+                )
+                expected = f(2π / 24 * h)
```
glennmoy, line 11: I was also suggesting you update these tests as well and hard code them.

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+    Periodic(f, period) -> Periodic
+
+A constructor for [`Periodic`](@ref).
+Returns a `Periodic` transform with zero phase shift.
+"""
+function Periodic(f, period)
+    return Periodic(f, period, zero(typeof(period)))
+end
```
glennmoy, line 34: could this be one line?

```
Periodic(f, period::T) where T = Periodic(f, period, zero(T))
```

src/periodic.jl
```diff
@@ -0,0 +1,95 @@
...
+    in the same units as the input. Increasing the value translates the function to the
+    right, toward higher/later input values.
+"""
+struct Periodic{T} <: Transform where T <: Union{Real, Period}
+    f::Union{typeof(cos), typeof(sin)}
+    period::T
+    phase_shift::T
+
+    function Periodic(f, period, phase_shift)
+        if period <= zero(typeof(period))
```
glennmoy, line 19:
```suggestion
        period <= zero(T) && throw(ArgumentError("period must be strictly positive."))
```

might also fit on one line

src/periodic.jl
glennmoy: yes - that's better idea

test/periodic.jl
bencottier: True. I think I'll have to change the name of `expected` when it is reassigned? E.g.

`expected = reshape(expected, (3, 2))`

test/periodic.jl
bencottier: It's not necessary here but I think it's important to test as high a precision as possible somewhere so that we know if it becomes less precise.

test/periodic.jl
glennmoy: or just call it `_expected`

test/periodic.jl
```diff
@@ -2,47 +2,32 @@
...
-                    f,
-                    ZonedDateTime(2015, 6, 1, tz"America/Winnipeg") + Hour(h),
-                    Day(1),
-                )
-                expected = f(2π / 24 * h)
-                @test result ≈ expected atol=1e-14
-            end
+            x = ZonedDateTime(2015, 6, 1, tz"America/Winnipeg") .+ collect(Hour.(0:2))
+            result = _periodic.(f, x, Day(1))
+            expected = Dict(sin => [0., 0.2588, 0.5], cos => [1.0, 0.9659, 0.866])[f]
```
glennmoy, line 7: easy to miss the `[f]` at the end here. 

I'd just put `@test result == expected[f] atol=1e-4`

test/periodic.jl
```diff
@@ -2,47 +2,32 @@
...
-            for h in 0:24
-                result = _periodic(
-                    f,
-                    ZonedDateTime(2015, 6, 1, tz"America/Winnipeg") + Hour(h),
-                    Day(1),
-                )
-                expected = f(2π / 24 * h)
-                @test result ≈ expected atol=1e-14
-            end
+            x = ZonedDateTime(2015, 6, 1, tz"America/Winnipeg") .+ collect(Hour.(0:2))
```
glennmoy, line 5: can we give these more meaningful names other than `x`? Like `normal_day`, `spring_dst`, `fall_dst`?

test/periodic.jl
```diff
@@ -2,47 +2,32 @@
...
+            @test result ≈ expected atol=1e-4
         end
 
         @testset "phase shift" begin
-            @test _periodic(sin, DateTime(1), Week(1)) == 0
-            @test _periodic(sin, DateTime(1), Week(1), Day(7)) == 0
-            @test _periodic(sin, DateTime(1), Week(1), Day(-7)) == 0
+            @test _periodic(sin, DateTime(2018), Week(1)) == 0
+            @test _periodic(sin, DateTime(2018), Week(1)) == 0
+            @test _periodic(sin, DateTime(2018), Week(1)) == 0
```
glennmoy, line 26: are these supposed to take `Day(7)`, `Day(-7)`?

glennmoy (approved):

src/periodic.jl
```diff
@@ -0,0 +1,93 @@
...
+    in the same units as the input. Increasing the value translates the function to the
+    right, toward higher/later input values.
+"""
+struct Periodic{T} <: Transform where T <: Union{Real, Period}
+    f::Union{typeof(cos), typeof(sin)}
+    period::T
+    phase_shift::T
+
+    function Periodic(f, period::T, phase_shift::T) where T
+        period <= zero(T) && throw(ArgumentError("period must be strictly positive."))
```
glennmoy, line 25:
```suggestion
        period > zero(T) || throw(ArgumentError("period must be strictly positive."))
```

I realise this is equivalent but perhaps uses fewer characters.

test/periodic.jl
glennmoy: also might help to put them in their own `@testsets`

test/periodic.jl
bencottier: Oops. Copy-paste fail.

src/periodic.jl
bencottier: It's cool I like how it reads as "this had better be positive OR I'll complain" 😄

src/periodic.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+    Periodic(f, period) -> Periodic
+
+A constructor for [`Periodic`](@ref).
+Returns a `Periodic` transform with zero phase shift.
+"""
+Periodic(f, period::T) where T = Periodic(f, period, zero(T))
+
+function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
+    x[:] = P.f.(2π .* (x .- P.phase_shift) / P.period)
```
glennmoy, line 39: > Increasing the value translates the function to the right, toward higher/later input values.

Is this correct? Wouldn't this translate it left?

glennmoy, line 39: I had to double check this was indeed true

https://www.wolframalpha.com/input/?i=plot+sin%28x%29+and+sin%28x-1%29

bencottier: I was just about to share a Desmos interactive plot :)

src/periodic.jl
```diff
@@ -0,0 +1,93 @@
...
+
+Computes the value of periodic function `f` at the given instant in time.
+
+# Arguments
+* `f`: the periodic function
+* `period`: the function period
+* `phase_shift`: adjusts the phase of the periodic function. Increasing the value translates
+    the function to the right, toward higher/later input values.
+"""
+function _periodic(f, instant, period, phase_shift=Day(0))
```
glennmoy, line 73: I've only just noticed that we use this for `TimeTypes` but not for reals... Do we know why this is? might be worth explaning in the docstring

glennmoy: I think it stems from the fact that the `instant`s we tend to pass in are far from the "origin" of time (`DateTime(0)`) and so `(x .- t) / period` is a huge number and calling `sin` / `cos` on this will inevitably introduce rounding errors even if `sin(0) - sin(2\pi * N) = 0` computationally this diverges as N grows large.

So to compensate we define an origin closer to the `instant` which, by virtue of periodicity, is equivalent to `DateTime(0)`.

src/periodic.jl
nicoleepp: A note on this, it's not trivial to construct a new table of the same type as the old one (aka the way to do this would be different for every table AFAIK) so just returning arrays helps with this too

src/periodic.jl
bencottier: I will leave handling this similarly for `Real` to another time, making an issue. We generally expect the `TimeType` inputs to be much further from the origin, but there's nothing stopping large inputs for `Real` either.

Issue #6, opened 2021-02-01
Temporal transformation


glennmoy:
Such as 
* hour of day (HoD)
* hour of week (HoW)
* etc

MWE
```julia
x = DateTime(2020, 1, 1, 9, 0):Hour(1):DateTime(2020, 5, 7, 16, 0)

# few ways we might define this
hod = Temporal(hour_of_day)
hod = HoD()

Transform.apply!(x, hod)
```

Note: as with #5 , a "custom transform" type may be preferable rather than defining a new transform for each function.

PR #11, opened 2021-02-03
Add HoD transform

nicoleepp:
Part of  #6 

TODO:

- [x] Support `inds` for AbstractArrays
- [x] Support Tables
- [x] Support `cols` for Tables
- [x] Re-add all the tests that I commeted out, clean up, remove packages that I added for local testing

### DIFF
--- a/src/temporal.jl
+++ b/src/temporal.jl
@@ -13,3 +13,14 @@ function apply(A::AbstractArray, t::HoD; dims=:, kwargs...)
 
     return [_apply(x, t; kwargs...) for x in eachslice(A, dims=dims)]
 end
+
+function apply(table, t::HoD; cols=nothing, kwargs...)
+    Tables.istable(table) || throw(MethodError(apply!, (table, t)))
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [_apply(getproperty(columntable, cname), t)  for cname in cnames]
+end
--- a/src/temporal.jl
+++ b/src/temporal.jl
@@ -1,23 +1,15 @@
 """
-    CustomTransform <: Transform
+    HoD <: Transform
 
-Apply the function passed in to CustomTransform as the transform.
+Get the hour of day corresponding to the data.
 """
 struct HoD <: Transform end
 
-function _apply!(x, ::HoD; kwargs...)
-    x[:] = hour.(x)
-    return x
+
+_apply(x, ::HoD) = hour.(x)
+
+function apply(A::AbstractArray, t::HoD; dims=:, kwargs...)
+    dims == Colon() && return _apply(A, t; kwargs...)
+
+    return [_apply(x, t; kwargs...) for x in eachslice(A, dims=dims)]
 end
-
-# struct HoW() <: Transform end
-
-# function _apply!(x, ::HoD; kwargs...)
-
-#     function hour_of_week(x)
-#         # TODO: this doesn't handle DST
-#         return dayofweek(x) * 24 + hour(x)
-#     end
-#     x[:] = hour_of_week.(x)
-#      return x
-# end
--- a/src/temporal.jl
+++ b/src/temporal.jl
@@ -1,0 +1,23 @@
+"""
+    CustomTransform <: Transform
+
+Apply the function passed in to CustomTransform as the transform.
+"""
+struct HoD <: Transform end
+
+function _apply!(x, ::HoD; kwargs...)
+    x[:] = hour.(x)
+    return x
+end
+
+# struct HoW() <: Transform end
+
+# function _apply!(x, ::HoD; kwargs...)
+
+#     function hour_of_week(x)
+#         # TODO: this doesn't handle DST
+#         return dayofweek(x) * 24 + hour(x)
+#     end
+#     x[:] = hour_of_week.(x)
+#      return x
+# end
--- a/src/temporal.jl
+++ b/src/temporal.jl
@@ -8,10 +8,16 @@ struct HoD <: Transform end
 
 _apply(x, ::HoD) = hour.(x)
 
-function apply(A::AbstractArray, t::HoD; dims=:, kwargs...)
-    dims == Colon() && return _apply(A, t; kwargs...)
+function apply(A::AbstractArray, t::HoD; dims=:, inds=:, kwargs...)
+    if dims === Colon()
+        if inds === Colon()
+            return _apply(A, t; kwargs...)
+        else
+            return [_apply(A[ind], t; kwargs...) for ind in inds]
+        end
+    end
 
-    return [_apply(x, t; kwargs...) for x in eachslice(A, dims=dims)]
+    return [_apply(x[inds], t; kwargs...) for x in eachslice(A, dims=dims)]
 end
 
 function apply(table, t::HoD; cols=nothing, kwargs...)
@@ -22,5 +28,5 @@ function apply(table, t::HoD; cols=nothing, kwargs...)
     columntable = Tables.columns(table)
 
     cnames = cols === nothing ? propertynames(columntable) : cols
-    return [_apply(getproperty(columntable, cname), t)  for cname in cnames]
+    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
 end
--- a/src/temporal.jl
+++ b/src/temporal.jl
@@ -7,26 +7,3 @@ struct HoD <: Transform end
 
 
 _apply(x, ::HoD) = hour.(x)
-
-function apply(A::AbstractArray, t::HoD; dims=:, inds=:, kwargs...)
-    if dims === Colon()
-        if inds === Colon()
-            return _apply(A, t; kwargs...)
-        else
-            return [_apply(A[ind], t; kwargs...) for ind in inds]
-        end
-    end
-
-    return [_apply(x[inds], t; kwargs...) for x in eachslice(A, dims=dims)]
-end
-
-function apply(table, t::HoD; cols=nothing, kwargs...)
-    Tables.istable(table) || throw(MethodError(apply!, (table, t)))
-
-    # Extract a columns iterator that we should be able to use to mutate the data.
-    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
-    columntable = Tables.columns(table)
-
-    cnames = cols === nothing ? propertynames(columntable) : cols
-    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
-end
--- a/src/transformers.jl
+++ b/src/transformers.jl
@@ -44,14 +44,12 @@ Does not need to be extended unless a mutating [`Transform`](@ref) is not possib
 function apply end
 
 """
-    apply!(A::AbstractArray{T}, ::Transform; dims=:, kwargs...) where T <: Real
+    apply!(A::AbstractArray, ::Transform; dims=:, kwargs...)
 
 Applies the [`Transform`](@ref) to each element of `A`.
 Optionally specify the `dims` to apply the [`Transform`](@ref) along certain dimensions.
 """
-function apply!(
-    A::AbstractArray{T}, t::Transform; dims=:, kwargs...
-) where T <: Real
+function apply!(A::AbstractArray, t::Transform; dims=:, kwargs...)
     dims == Colon() && return _apply!(A, t; kwargs...)
 
     for x in eachslice(A; dims=dims)
--- a/src/transformers.jl
+++ b/src/transformers.jl
@@ -28,20 +28,73 @@ Non-mutating version of [`transform!`](@ref).
 function transform end
 
 """
-    Transforms.apply!(data::T, ::Transform; kwargs...) -> T
+    apply(data::T, ::Transform; kwargs...) -> T
 
-Applies the [`Transform`](@ref) mutating the input `data`.
+Applies the [`Transform`](@ref) to the data. New transforms should usually only extend
+`_apply` which this method delegates to.
+
+Where possible, this should be extended for new data types `T`.
+"""
+function apply end
+
+"""
+    apply!(data::T, ::Transform; kwargs...) -> T
+
+Applies the [`Transform`](@ref) mutating the input `data`. New transforms should usually
+only extend `_apply!` which this method delegates to.
+
 Where possible, this should be extended for new data types `T`.
 """
 function apply! end
 
-"""
-    Transforms.apply(data::T, ::Transform; kwargs...) -> T
 
-Non-mutating version of [`apply!`](@ref), which it delegates to by default.
-Does not need to be extended unless a mutating [`Transform`](@ref) is not possible.
 """
-function apply end
+    apply(A::AbstractArray, ::Transform; dims=:, inds=:, kwargs...)
+
+Applies the [`Transform`](@ref) to each element of `A`.
+
+Optionally specify the `dims` to apply the [`Transform`](@ref) along certain dimensions
+and `inds` will be the indices to apply the Transform to along the `dims` specified.
+
+If `dims` === : (all dimensions), then `inds` will be the global indices of the array,
+instead of being relative to a certain dimension.
+"""
+function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
+    if dims === Colon()
+        if inds === Colon()
+            return _apply(A, t; kwargs...)
+        else
+            if A isa KeyedArray
+                # KeyedArrays don't support indexing into them via an array of indices
+                return [_apply(A[ind], t; kwargs...) for ind in inds]
+            else
+                # Apply to global indices `inds`, not `inds` relative to a certain dimension
+                return _apply(A[inds], t; kwargs...)
+            end
+        end
+    end
+
+    return mapslices(x -> _apply(x[inds], t; kwargs...), A, dims=dims)
+end
+
+"""
+    apply(table, ::Transform; cols=nothing, kwargs...)::Vector
+
+Applies the [`Transform`](@ref) to each of the specified columns in the `table`.
+If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
+"""
+function apply(table, t::Transform; cols=nothing, kwargs...)
+    Tables.istable(table) || throw(MethodError(apply!, (table, t)))
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
+end
+
+_apply(x, t::Transform; kwargs...) = apply!(_try_copy(x), t; kwargs...)
 
 """
     apply!(A::AbstractArray, ::Transform; dims=:, kwargs...)
@@ -59,10 +112,8 @@ function apply!(A::AbstractArray, t::Transform; dims=:, kwargs...)
     return A
 end
 
-apply(x, t::Transform; kwargs...) = apply!(_try_copy(x), t; kwargs...)
-
 """
-    Transforms.apply!(table::T, ::Transform; cols=nothing)::T where T
+    apply!(table::T, ::Transform; cols=nothing)::T where T
 
 Applies the [`Transform`](@ref) to each of the specified columns in the `table`.
 If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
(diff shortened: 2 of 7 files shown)


src/temporal.jl
```diff
@@ -0,0 +1,23 @@
+"""
+    CustomTransform <: Transform
+
+Apply the function passed in to CustomTransform as the transform.
+"""
+struct HoD <: Transform end
+
+function _apply!(x, ::HoD; kwargs...)
+    x[:] = hour.(x)
```
nicoleepp, line 9: Does this even make sense as mutating? Tests currently error on the conversion from a `DateTime` array to now being one that would hold `Int`s

rofinn: Yeah, I don't think mutating make sense here.

src/transformers.jl
```diff
@@ -44,14 +44,12 @@ Does not need to be extended unless a mutating [`Transform`](@ref) is not possib
...
-    apply!(A::AbstractArray{T}, ::Transform; dims=:, kwargs...) where T <: Real
+    apply!(A::AbstractArray, ::Transform; dims=:, kwargs...)
 
 Applies the [`Transform`](@ref) to each element of `A`.
 Optionally specify the `dims` to apply the [`Transform`](@ref) along certain dimensions.
 """
-function apply!(
-    A::AbstractArray{T}, t::Transform; dims=:, kwargs...
-) where T <: Real
+function apply!(A::AbstractArray, t::Transform; dims=:, kwargs...)
```
bencottier, line 52: I don't think this change is necessary given we can't mutate arrays for HoD, but maybe it's right in general. I don't know the reason it was restricted to `Real`.

src/temporal.jl
bencottier: For that reason the set of methods for `HoD` could end up similar to `Periodic` transform with `TimeType` input.

bencottier: Aside: I made an issue about how to transform Tables for cases like this https://github.com/invenia/Transforms.jl/issues/12

src/transformers.jl
nicoleepp: @glennmoy ?

src/temporal.jl
```diff
@@ -0,0 +1,23 @@
+"""
+    CustomTransform <: Transform
```
glennmoy, line 2: is this a typo or are you planning on implementing it as part of HoD?

src/transformers.jl
glennmoy: I was following Impute.jl's example.

Removing the type restriction seems ok to me.

We might still like to apply mutating transforms to non-numerical arrays, e.g. string, ZonedDateTimes, even though the ones we have currently implemented can't do that.

src/temporal.jl
nicoleepp: typo

test/temporal.jl
```diff
@@ -0,0 +1,312 @@
...
+                @test Transforms.apply(M, hod; dims=d) == expected
+                @test hod(M; dims=d) == expected
+
+                # Test the tranform was not mutating
+                @test M != expected
+            end
+
+            @testset "dims = 1" begin
+                d = 1
+                expected = [[1, 9], [2, 10], [3, 11]]
```
glennmoy, line 88: I guess this is an example of a return being different to it's input? Matrix vs Vector{Vector}?

test/temporal.jl
```diff
@@ -0,0 +1,312 @@
...
+            @testset "cols = all" begin
+                cols = [:a, :b]
+                @test Transforms.apply(nt, hod; cols=cols) == expected
+                @test hod(nt; cols=cols) == expected
+            end
+
+            @testset "cols = :a" begin
+                cols = [:a]
+                # TODO: this is kind of ugly if only a single Vector is expected
+                expected = [[0, 1, 2]]
```
glennmoy, line 256: and this being another weird/ugly example.

test/temporal.jl
glennmoy: We could `cat` the result and keep the input/output consistent, but `eachslice` doesn't preserve orientation, e.g.

```julia

julia> r = [x .+ 1 for x in eachslice(A, dims=1)]
3-element Array{Array{Int64,1},1}:
 [2, 3, 4]
 [5, 6, 7]
 [8, 9, 10]

julia> s = [x .+ 1 for x in eachslice(A, dims=2)]
3-element Array{Array{Int64,1},1}:
 [2, 5, 8]
 [3, 6, 9]
 [4, 7, 10]
```

we could instead replace it with `mapslices` for which 
>The results are concatenated along the remaining dimensions

```julia
julia> t = mapslices(x->x.+1, A, dims=1)
3×3 Array{Int64,2}:
 2  3   4
 5  6   7
 8  9  10

julia> t = mapslices(x->x.+1, A, dims=2)
3×3 Array{Int64,2}:
 2  3   4
 5  6   7
 8  9  10
```

glennmoy: I think doing something like this would work
```julia
julia> function apply(A::AbstractArray, t::HoD; dims=:, inds=:, kwargs...)
           if dims === Colon()
               if inds === Colon()
                   return _apply(A, t; kwargs...)
               else
                   return _apply(A[inds], t; kwargs...)
               end
           end

           return mapslices(x->_apply(x[inds], t; kwargs...), A, dims=dims)
       end
apply (generic function with 1 method)

julia>

julia>

julia> apply(M, HoD(), dims=1)
3×2 Array{Int64,2}:
 1   9
 2  10
 3  11

julia> apply(M, HoD(), dims=2)
3×2 Array{Int64,2}:
 1   9
 2  10
 3  11
```

src/temporal.jl
```diff
@@ -0,0 +1,32 @@
+"""
+    HoD <: Transform
```
glennmoy, line 2: when adding DoW or any other temporal transform later it might help to define a supertype `TemporalTransform` - just an idea

src/temporal.jl
```diff
@@ -0,0 +1,32 @@
...
+
+function apply(table, t::HoD; cols=nothing, kwargs...)
+    Tables.istable(table) || throw(MethodError(apply!, (table, t)))
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
```
bencottier, line 31: Should we have `kwargs` here? It's not currently used by `_apply`.

src/temporal.jl
```diff
@@ -0,0 +1,32 @@
...
+            return _apply(A, t; kwargs...)
+        else
+            return [_apply(A[ind], t; kwargs...) for ind in inds]
+        end
+    end
+
+    return [_apply(x[inds], t; kwargs...) for x in eachslice(A, dims=dims)]
+end
+
+function apply(table, t::HoD; cols=nothing, kwargs...)
```
bencottier, line 23: We should add a docstring including the return type.

src/temporal.jl
```diff
@@ -0,0 +1,32 @@
...
+    HoD <: Transform
+
+Get the hour of day corresponding to the data.
+"""
+struct HoD <: Transform end
+
+
+_apply(x, ::HoD) = hour.(x)
+
+function apply(A::AbstractArray, t::HoD; dims=:, inds=:, kwargs...)
```
bencottier, line 11: Do we need `dims` at all for this transform? I think it doesn't matter because it's one-to-one.

bencottier: Ah, looks like it affects returning Matrix vs. vector of vectors. In that case I'm not sure - I don't know if there's a use case for having the result sliced up in different ways. Currently `Periodic` doesn't use`dims`.

test/temporal.jl
bencottier: I made a different thread about this but if we `cat` the result, for this transform, why do we need `dims`? As per your example the result is the same. One reason is this method might become common and generalised in `transformers.jl`?

glennmoy: probably we don't need dims then - and it should be absorbed in the `kwargs...`

src/temporal.jl
glennmoy: >I don't know if there's a use case for having the result sliced up in different ways

I don't either, it's a consequence of `eachslice` not that we particularly want it that way.

test/temporal.jl
nicoleepp: Yeah I was aiming on having this apply method be the default, and moving it to transforms.jl if it's cleared as being a good general case

src/temporal.jl
nicoleepp: I will note `dims` doesn't make a difference unless `inds` is passed in too for this transform with the updated implementation following this round of code review. 

Also, I'm trying to make this apply method be the canonical apply method that gets used for non-mutating transforms.

src/temporal.jl
nicoleepp: yeah, out of scope for this MR though until we find more temporal transforms

src/transformers.jl
```diff
@@ -28,30 +28,81 @@ Non-mutating version of [`transform!`](@ref).
...
     dims == Colon() && return _apply!(A, t; kwargs...)
 
     for x in eachslice(A; dims=dims)
```
bencottier, line 108: `mapslices` here too?

bencottier (approved):
I have a couple questions and suggestions but this looks good to me.

We've agreed in the past to return `AbstractArray` for non-mutating `apply`, for consistency between `Transform`s. I'm still uncertain whether that's best, but I think we should leave that to further discussion outside this PR.

src/transformers.jl
```diff
@@ -28,20 +28,73 @@ Non-mutating version of [`transform!`](@ref).
...
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
+end
+
+_apply(x, t::Transform; kwargs...) = apply!(_try_copy(x), t; kwargs...)
```
glennmoy, line 97:
```suggestion
_apply(x, t::Transform; kwargs...) = _apply!(_try_copy(x), t; kwargs...)
```

test/power.jl
```diff
@@ -30,75 +30,79 @@
...
         A = AxisArray([1 2 3; 4 5 6], foo=["a", "b"], bar=["x", "y", "z"])
-        expected = AxisArray([1 8 27; 64 125 216], foo=["a", "b"], bar=["x", "y", "z"])
+        expected = [1 8 27; 64 125 216]
+        axisarray_expected = AxisArray([1 8 27; 64 125 216], foo=["a", "b"], bar=["x", "y", "z"])
 
         @testset "dims = $d" for d in (Colon(), 1, 2)
-            transformed = Transforms.apply(A, p; dims=d)
-            @test transformed isa AxisArray
-            @test transformed == expected
+            @test Transforms.apply(A, p; dims=d) == expected
```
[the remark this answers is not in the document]
glennmoy: Because we haven't asked `apply` to return a KeyedArray - just an array.

test/temporal.jl
```diff
@@ -53,10 +42,19 @@
...
-                @test Transforms.apply(x, hod; dims=d, inds=[1]) == expected
+                @test Transforms.apply(x, hod; dims=d, inds=[2, 3, 4, 5]) == expected
             end
         end
+
+        @testset "DST" begin
+            x = ZonedDateTime(2020, 3, 7, 9, 0, tz"America/New_York"):Hour(1):ZonedDateTime(2020, 3, 8, 9, 0, tz"America/New_York")
+
+            # expected result skips the DST transition hour of 2
+            expected = [9:23..., 0, 1, 3:9...]
```
glennmoy, line 53:
```suggestion
            expected_dst= [9:23..., 0, 1, 3:9...]
```
only because it could get interpreted that we've reassigned the `expected` used in a bunch of other tests.

test/temporal.jl
```diff
@@ -163,14 +164,20 @@
...
+
+                expected = [1 9 10; 2 10 11]
+                @test Transforms.apply(M, hod; dims=d, inds=[1, 2]) == expected
             end
 
             @testset "dims = 2" begin
                 d = 2
-                expected = [[2], [10], [11]]
-                @test Transforms.apply(A, hod; dims=d, inds=[2]) == expected
+                expected = [9 10; 10 11]
```
glennmoy, line 176: this is very different to the answer above - in spite of the different `inds`. What caused this?

src/transformers.jl
```diff
@@ -28,20 +28,73 @@ Non-mutating version of [`transform!`](@ref).
...
+instead of being relative to a certain dimension.
 """
-function apply end
+function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
+    if dims === Colon()
+        if inds === Colon()
+            return _apply(A, t; kwargs...)
+        else
+            if A isa KeyedArray
+                # KeyedArrays don't support indexing into them via an array of indices
```
glennmoy, line 68: what about `A[:][inds]`?

```julia
julia> A = KeyedArray(rand(3, 3); node=["a", "b", "c"], time=[1, 2, 3]);

julia> inds=[2, 3];

julia> A[inds]
ERROR: MethodError: no method matching keys_getindex(::Tuple{Array{Int64,1}}, ::Tuple{})
Closest candidates are:
  keys_getindex(::Tuple{}, ::Tuple{}) at /Users/glenn/.julia/packages/AxisKeys/iJ0Vf/src/struct.jl:122
  keys_getindex(::Tuple, ::Tuple{Integer,Vararg{Any,N} where N}) at /Users/glenn/.julia/packages/AxisKeys/iJ0Vf/src/struct.jl:95
  keys_getindex(::Tuple, ::Tuple{Base.Slice,Vararg{Any,N} where N}) at /Users/glenn/.julia/packages/AxisKeys/iJ0Vf/src/struct.jl:99
  ...
Stacktrace:
 [1] keys_getindex at /Users/glenn/.julia/packages/AxisKeys/iJ0Vf/src/struct.jl:109 [inlined]
 [2] getindex(::KeyedArray{Float64,2,NamedDimsArray{(:node, :time),Float64,2,Array{Float64,2}},Tuple{Array{String,1},Array{Int64,1}}}, ::Array{Int64,1}) at /Users/glenn/.julia/packages/AxisKeys/iJ0Vf/src/struct.jl:80
 [3] top-level scope at REPL[34]:1

julia> A[:][inds]
2-element NamedDimsArray(::Array{Float64,1}, (:node,)):
↓ node  0.8601795204696292
        0.10124610540366952

julia> M[:][inds]
2-element Array{Int64,1}:
 4
 7
```

maybe this is less memory efficient? but `A[:]` would be allocating a new array anyway.

Also means we can get rid of the dependency on `AxisKeys`.

src/transformers.jl
```diff
@@ -28,20 +28,73 @@ Non-mutating version of [`transform!`](@ref).
...
+                return [_apply(A[ind], t; kwargs...) for ind in inds]
+            else
+                # Apply to global indices `inds`, not `inds` relative to a certain dimension
```
glennmoy, line 71: Just working this through myself. 

`inds` with `dim=:`, vectorises the elements of an array

So `apply(M, t; inds=[2,3])` yields

```julia
julia> M = [1 2 3; 4 5 6; 7 8 9];

julia> inds = [2, 3];

julia> M[inds]
2-element Array{Int64,1}:
 4
 7
```

Is this an intuitive result?
From our POV it is because we see what `dims=:` is doing. 
But will a new user have the same understanding?

Is it better to default `dims=1` instead? so that `apply(M, t; inds=[2,3])` gives
```julia
julia> M[:, inds]
3×2 Array{Int64,2}:
 2  3
 5  6
 8  9
```

I'm not suggesting this way is better, I'm fine with this way too.
 I'm just trying to imagine what a non-developer would think when trying `apply(M, t; inds=[2,3])` for the first time.

I guess the docs/examples we can show a user what to expect.

src/temporal.jl
glennmoy: >Also, I'm trying to make this apply method be the canonical apply method that gets used for non-mutating transforms.

What was the reason this last commit was needed? It seemed out of scope for me but maybe I missed something?

test/power.jl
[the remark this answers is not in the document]
bencottier: This is for `AxisArray`. If I follow `apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)` then I think it does `_apply` (without wrapping in an array), which I thought would preserve the type. But maybe this relates to your suggestion? i.e. it's currently calling `apply!` when it should be `_apply!` https://github.com/invenia/Transforms.jl/pull/11/files#r573720397

test/temporal.jl
bencottier: Not sure what you mean by answer above, but I'll explain how I checked this.

```
1  9   10
2  10  11
```

`dims=2` slices into rows `[1, 9, 10]` and `[2, 10, 11]`. `inds = [2, 3]` picks the 2nd and 3rd element of each of those rows which is `9, 10` and `10, 11` respectively.

I thought this same logic checked out for the other test cases but let me know if you think otherwise, or it's not how you expect it to work.

src/temporal.jl
bencottier: > What was the reason this last commit was needed? It seemed out of scope for me but maybe I missed something?

Fair to say it's out of scope. Maybe it should be split up.

src/transformers.jl
nicoleepp: This doesn't work for vectors

src/transformers.jl
nicoleepp: I don't really have any pre-existing intuition around `dims` or operations on matrices so I can't really say

src/temporal.jl
nicoleepp: I wanted to be more clear on the general case, how we want `apply` to work with dims and inds instead of getting lost in the details of hod. Once the general case is clear the individual transform cases become way more trivial

src/transformers.jl
glennmoy: >This doesn't work for vectors

what MWE did you use?

```julia
julia> v = [1, 2, 3, 4, 5];

julia> inds=[2, 3];

julia> v[:][inds]
2-element Array{Int64,1}:
 2
 3
```

src/temporal.jl
glennmoy: fair enough

test/power.jl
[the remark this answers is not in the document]
glennmoy: >This is for AxisArray

ah my mistake. Then this is `AxisArrays` being annoying and not conserving type when operating on them

```julia
julia> A = AxisArray([1 2 3; 4 5 6], foo=["a", "b"], bar=["x", "y", "z"]);

julia> A .+ 1
2×3 Array{Int64,2}:
 2  3  4
 5  6  7
```

test/temporal.jl
glennmoy: >Not sure what you mean by "answer above",

Sorry, I meant in relation to the expected answer it's replacing: `expected = [[2], [10], [11]]`

Before this commit, for `d=2`, `inds=[2]` the result was `[[2], [10], [11]]` a `Vector{Vector}`

Now, for `d=2`, `inds=[2, 3]` the result is `[9 10; 10 11]` an `Array{Int, 2}`.

It's correct, but I'm wondering why it was so different before. I'm guessing it's the result of now using `mapslices`?

src/transformers.jl
glennmoy: Cool. Let's leave it as is.

src/transformers.jl
```diff
@@ -28,30 +28,81 @@ Non-mutating version of [`transform!`](@ref).
...
+                return _apply(A[inds], t; kwargs...)
+            end
+        end
+    end
+
+    return mapslices(x -> _apply(x[inds], t; kwargs...), A, dims=dims)
```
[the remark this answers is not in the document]
nicoleepp: mapslices `x` is not a KeyedArray, it's a view of the array

test/temporal.jl
bencottier: Ah, I was confused because I was looking at the full diff.

nicoleepp: The dims were in the opposite order before thanks to eachslice versus mapslices differences

src/temporal.jl
nicoleepp: https://github.com/invenia/Transforms.jl/pull/15

src/transformers.jl
nicoleepp: This works just as welll since it's mutating the array, I briefly tried using mapslices and things were breaking so I'd opt to leave as is. It does return the result we expect

src/transformers.jl
nicoleepp: Nvm, whatever example it was it seems to be working now

nicoleepp:
Closing in favour of https://github.com/invenia/Transforms.jl/pull/16

Issue #12, opened 2021-02-04
Use a consistent interface for Tables where input type cannot be converted


bencottier:
I'm uncertain about this. There's a more general question of what operations on Tables we want to provide, vs. leaving to the user.

Currently, we have an [`apply!` for tables](https://github.com/invenia/Transforms.jl/blob/3fe2b2718b262f99e8ac215e70e01e041fdad3ba/src/transformers.jl#L72-L86) that mutates the table columns in-place.

However this doesn't work when the input type can't be converted to the output type, e.g. from DateTime to Int for the hour-of-day transform. We could replace the column with the new type, but users should have a choice whether to do such a "strong" mutation.

Alternatively we could have an `apply!` that appends the new columns to the Table.

We could also have an `apply` that  returns the transformed columns on their own ([example](https://github.com/invenia/Transforms.jl/blob/7b2e35aebe587b858eec7b79e95be64704a513db/src/periodic.jl#L49-L55)), which gives users the most flexibility.

nicoleepp:
I don't like

> Alternatively we could have an apply! that appends the new columns to the Table.

just because `apply!` done on Arrays would not have the same behaviour and that's confusing

For that reason I think

> We could also have an apply that returns the transformed columns on their own (example), which gives users the most flexibility.

is the best option, and let the user append the result to the table explicitly

nicoleepp:
If we really wanted

> Alternatively we could have an apply! that appends the new columns to the Table.

It should probably be a third function, like `apply_append!(x::Table, ::Transfrom; newcolnames = [:a. :b], kwargs...)`

or

`apply!(x::Table, ::Transform; append=false, newcolnames=[:a, :b], kwargs...)` but this last one might be confusing as sometimes it would mutate the cols that exist and sometimes append

glennmoy:
> is the best option, and let the user append the result to the table explicitly

It also gives them the flexibility to name it accordingly. Otherwise we'd have to come up with a naming convention in the package ☠️

PR #15, opened 2021-02-10
Implement general case non-mutating apply

nicoleepp:
Part of #12

### DIFF
--- a/src/periodic.jl
+++ b/src/periodic.jl
@@ -44,7 +44,7 @@ function _apply(x, P::Periodic{T}; kwargs...) where T <: Period
 end
 
 function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
-    x[:] = apply(x, P; kwargs...)
+    x[:] = _apply(x, P; kwargs...)
     return x
 end
 
--- a/src/periodic.jl
+++ b/src/periodic.jl
@@ -35,34 +35,17 @@ Returns a `Periodic` transform with zero phase shift.
 """
 Periodic(f, period::T) where T = Periodic(f, period, zero(T))
 
+function _apply(x, P::Periodic; kwargs...)
+    return P.f.(2π .* (x .- P.phase_shift) / P.period)
+end
+
+function _apply(x, P::Periodic{T}; kwargs...) where T <: Period
+    map(xi -> _periodic(P.f, xi, P.period, P.phase_shift), x)
+end
+
 function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
-    x[:] = P.f.(2π .* (x .- P.phase_shift) / P.period)
+    x[:] = apply(x, P; kwargs...)
     return x
-end
-
-# `U <: Period` needed to avoid method ambiguity with
-# `apply(table, P::Periodic{T}; cols=nothing) where T <: Period`
-function apply(
-    x::AbstractArray{T},
-    P::Periodic{U};
-    kwargs...
-) where {T <: TimeType, U <: Period}
-    return map(xi -> _periodic(P.f, xi, P.period, P.phase_shift), x)
-end
-
-"""
-    Transforms.apply(table, ::Periodic{T}; cols=nothing) where T <: Period -> Array
-
-Applies [`Periodic`](@ref) to each of the specified columns in `table`.
-If no `cols` are specified, then [`Periodic`](@ref) is applied to all columns.
-Returns an array containing each transformed column, in the same order as `cols`.
-"""
-function apply(table, P::Periodic{T}; cols=nothing) where T <: Period
-    Tables.istable(table) || throw(MethodError(apply, (table, P)))
-
-    columntable = Tables.columns(table)
-    cnames = cols === nothing ? propertynames(columntable) : cols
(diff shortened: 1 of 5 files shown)


src/transformers.jl
```diff
@@ -28,30 +28,80 @@ Non-mutating version of [`transform!`](@ref).
...
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    return [_apply(getproperty(columntable, cname), t; kwargs...)  for cname in cnames]
+end
+
+#TODO: should this be apply! ?
```
nicoleepp, line 94: I'm not sure what the difference is from calling `apply!` here instead of `_apply!`. 

The general `apply!` still calls `_apply!` so I think it just performs unneeded checks again? I may be wrong

test/periodic.jl
```diff
@@ -271,9 +269,12 @@
 
                 @testset "dims = $d" for d in (Colon(), 1, 2)
                     transformed = Transforms.apply(A, p; dims=d)
-                    @test transformed isa AxisArray
+                    # AxisArray doesn't preserve type when operations are performed on it
+                    @test transformed isa AbstractArray
                     @test transformed ≈ expected atol=1e-14
                 end
+
+                # TODO: confirm we don't support mutating periodic for Time Types
```
nicoleepp, line 277: @bencottier ?

src/transformers.jl
bencottier: Yeah, I think it should be `_apply!` rather than `apply!`.

However, I'm thinking about the move towards `apply` being more standard than `apply!`. The way things are, `_apply!` must be implemented for each transform.

If we want to switch things around, we could have a default `_apply!` similar to what you added in `periodic.jl`:

```
function _apply!(x, t::Transform; kwargs...)
    x[:] = _apply(x, t; kwargs...)
    return x
end
```

Not essential for this PR though.

test/periodic.jl
bencottier: Correct

bencottier (commented):
Looks good, I just have a few suggestions and uncertainties.

src/periodic.jl
```diff
@@ -35,34 +35,17 @@ Returns a `Periodic` transform with zero phase shift.
...
-Returns an array containing each transformed column, in the same order as `cols`.
-"""
-function apply(table, P::Periodic{T}; cols=nothing) where T <: Period
-    Tables.istable(table) || throw(MethodError(apply, (table, P)))
-
-    columntable = Tables.columns(table)
-    cnames = cols === nothing ? propertynames(columntable) : cols
-    return [apply(getproperty(columntable, cname), P) for cname in cnames]
+function _apply!(x::AbstractArray{T}, P::Periodic; kwargs...) where T <: Real
+    x[:] = apply(x, P; kwargs...)
```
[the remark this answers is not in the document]
nicoleepp: Oh yes, this was a mistake

src/transformers.jl
nicoleepp: Maybe the error if `_apply!` is not implemented is more clear though? Because not all transforms support non-mutating, and the above example will result in Conversion errors. Is a method not implemented error more clear than a conversion error?

src/transformers.jl
```diff
@@ -28,30 +28,80 @@ Non-mutating version of [`transform!`](@ref).
...
+Will return the same datatype except for AxisArrays as operations on those do not preserve
+type.
 """
-function apply end
+function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
+    if dims === Colon()
+        if inds === Colon()
+            return _apply(A, t; kwargs...)
+        else
+            return _apply(A[:][inds], t; kwargs...)
```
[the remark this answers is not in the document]
nicoleepp: I don't think we should not support Int inds, I will modify the comment to say the return type is not guaranteed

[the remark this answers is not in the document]
bencottier: Fair enough, the Int inds result is not that surprising if one knows it's a global index.

src/transformers.jl
bencottier: Hmm I see where you're coming from. Method not implemented error does seem clearer - "we don't support mutating for this transform" rather than "we tried and failed to support mutating for this transform".

bencottier (approved):

src/transformers.jl
glennmoy: Yeah I think it should be:
```
_apply(x, t::Transform; kwargs...) = _apply!(_try_copy(x), t; kwargs...)
```

The thinking was that you would only have to implement the mutating `_apply!` version for each new Transform.
Then for the non-mutating call to `_apply`, it should just copy the input and call `_apply!`.

So this was supposed to be a convenience for not having to define `_apply` for every new transform.

>The way things are, _apply! must be implemented for each transform.

yeah our philosophy has since changed to "Every transform must implement `_apply` and, where possible, `_apply!`, wherein `_apply` can just call that after copying the data".

So I think this  can still be suported?

PR #16, opened 2021-02-11
Add HoD transform

nicoleepp:
Part of #6 

Supersedes #11

### DIFF
--- a/test/temporal.jl
+++ b/test/temporal.jl
@@ -1,0 +1,173 @@
+@testset "temporal" begin
+
+    hod = HoD()
+    @test hod isa Transform
+
+    @testset "Vector" begin
+        x = collect(DateTime(2020, 1, 1, 9, 0):Hour(1):DateTime(2020, 5, 7, 9, 0))
+        # Expected result is an hour a day starting and ending on the 9th hour inclusive,
+        # with 126 full days in the middle
+        expected = [9:23..., repeat(0:23, 126)..., 0:9...]
+
+        @test Transforms.apply(x, hod) == expected
+        @test hod(x) == expected
+
+        # Test the tranform was not mutating
+        @test x != expected
+
+        @testset "StepRange" begin
+            x = DateTime(2020, 1, 1, 9, 0):Hour(1):DateTime(2020, 5, 7, 9, 0)
+            @test Transforms.apply(x, hod) == expected
+            @test hod(x) == expected
+        end
+
+        @testset "dims = $d" for d in (Colon(), 1)
+            @test Transforms.apply(x, hod; dims=d) == expected
+            @test hod(x; dims=d) == expected
+        end
+
+        @test_throws BoundsError Transforms.apply(x, hod; dims=2)
+
+        @testset "inds" begin
+            @test Transforms.apply(x, hod; inds=2:5) == expected[2:5]
+            @test Transforms.apply(x, hod; dims=:) == expected
+            @test Transforms.apply(x, hod; dims=1) == expected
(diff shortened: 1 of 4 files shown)


test/temporal.jl
```diff
@@ -0,0 +1,173 @@
...
+            # AxisArray doesn't preserve the type it operates on
+            @test transformed isa AbstractArray
+            @test transformed == expected
+        end
+
+        @testset "inds" begin
+            @test Transforms.apply(A, hod; inds=[2, 3]) == expected[[2, 3]]
+            @test Transforms.apply(A, hod; dims=:, inds=[2, 3]) == expected[[2, 3]]
+            @test Transforms.apply(A, hod; dims=1, inds=[2]) == reshape(expected[[2, 4, 6]], 1, 3)
+            @test Transforms.apply(A, hod; dims=2, inds=[2]) == reshape(expected[[3, 5]], 2, 1)
```
bencottier, line 100:
```suggestion
            @test Transforms.apply(A, hod; dims=2, inds=[2]) == reshape(expected[[3, 4]], 2, 1)
```
If I understand correctly, `dims=2, inds=[2]` is selecting the middle column, which is global index 3 and 4. So I think this passed by coincidence of global index 4 and 5 having the same value?

test/temporal.jl
```diff
@@ -0,0 +1,173 @@
...
+            transformed = Transforms.apply(A, hod; dims=d)
+            @test transformed isa KeyedArray
+            @test transformed == expected
+        end
+
+        @testset "inds" begin
+            @test Transforms.apply(A, hod; inds=[2, 3]) == expected[[2, 3]]
+            @test Transforms.apply(A, hod; dims=:, inds=[2, 3]) == expected[[2, 3]]
+            @test Transforms.apply(A, hod; dims=1, inds=[2]) == reshape(expected[[2, 4, 6]], 1, 3)
+            @test Transforms.apply(A, hod; dims=2, inds=[2]) == reshape(expected[[3, 5]], 2, 1)
```
bencottier, line 127:
```suggestion
            @test Transforms.apply(A, hod; dims=2, inds=[2]) == reshape(expected[[3, 4]], 2, 1)
```
As for AxisArray

bencottier (approved):
Looks good to me - just spotted a couple typos in the tests.

test/temporal.jl
```diff
@@ -0,0 +1,173 @@
...
+
+    @testset "AxisArray" begin
+        x = [
+            DateTime(2020, 1, 1, 1, 0):Hour(1):DateTime(2020, 1, 1, 2, 0);
+            DateTime(2020, 1, 1, 9, 0):Hour(1):DateTime(2020, 1, 1, 10, 0);
+            DateTime(2020, 1, 1, 10, 0):Hour(1):DateTime(2020, 1, 1, 11, 0)
+        ]
+        M = reshape(x, 2, 3)
+        A = AxisArray(M, foo=["a", "b"], bar=["x", "y", "z"])
+        expected = [1 9 10; 2 10 11]
```
glennmoy, line 84:
```suggestion
        expected = [1 9 11; 2 10 12]
```

want each element to be unique so we're sure `inds` is getting the right ones

test/temporal.jl
glennmoy: I think in general I'd prefer to write the expected result more explicitly (where it's not trivial) rather rely on `reshape` and indexing.

Instead of `reshape(expected[[3, 4]], 2, 1)` say `[9, 10]` 

It's more clear what the expected answer is.

glennmoy (approved):

PR #17, opened 2021-02-12
Add MeanStdScaling transform

### LINKED ISSUE #3: Normalisation / Scaling transform
A transform that  scales data according to some distribution properties

Since this is a stateful transform, we may also want to apply reverse it later and/or apply the same scaling but to different data.

There might be a few ways to do this but one I'm considering is that the `Scaling` transform is parameterised by the type of Distribution we want to [`fit`](https://juliastats.org/Distributions.jl/latest/fit/#Distribution-Fitting) to the data?

MWE
```julia
x = rand(100)
y = rand(10)

# an uninitialised Scaling Transform
gaussian = Scaling{Normal}()

# scales x according to normal distribution and initialises `gaussian.d` with the mean/std of x
Transform.apply!(x, gaussian)  

# applies `gaussian` to `y` using stored mean/std
Transform.apply!(y, gaussian) 

# inverts the `gaussian` transform restoring `x` to original state
Transform.apply(x, gaussian; inverse=true)  
```

Something similar is kind of done in [StatsBase](https://juliastats.org/StatsBase.jl/stable/transformations/#) but only for `ZScore` and `UnitRange`

bencottier:
Closes #3

Note: the notion of `dims` is now consistent with `mapslices` and `Statistics.mean/std` rather than `eachslice`. To do that, I inverted the dims passed to the general `apply!` method for arrays. `Statistics.mean/std` was used in `FeatureEngineering`. Related issue: #18


src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+    MeanStdScaling() = new(Dict(), Dict())
+end
+
+"""
+    MeanStdScaling(data; kwargs...) <: Scaling
+
+Construct a [`MeanStdScaling`](@ref) using the mean and standard deviation of the data.
+
+# Keyword arguments
+*
```
bencottier, line 47: Was thinking of explaining `dims` for arrays and `cols` for tables, from `populate_stats!`, here.

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+# Convenience method
+IdentityScaling(args...; kwargs...) = IdentityScaling()
+
+apply!(A::AbstractArray{T}, scaling::IdentityScaling; kwargs...) where T <: Real = A
+
+"""
+    MeanStdScaling(mean, std) <: Scaling
+
+Linearly scale the data by a statistical `mean` and standard deviation `std`.
+The first `mean` and `std` calculated on some data will remain fixed.
```
bencottier, line 23: Important to take note of this throughout.

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+"""
+    IdentityScaling
+
+Represents the no-op scaling which simply returns the data it is applied on.
+"""
+struct IdentityScaling <: Scaling end
+# Convenience method
+IdentityScaling(args...; kwargs...) = IdentityScaling()
+
+apply!(A::AbstractArray{T}, scaling::IdentityScaling; kwargs...) where T <: Real = A
```
glennmoy, line 17:
```suggestion
apply!(A, ::IdentityScaling; kwargs...) = A
```

src/scaling.jl
glennmoy: I think this could be better phrased?

```
Once computed, the statistics of a `MeanStdScaling` are immutable.
```

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+    # Is this going too far to generalise between Arrays and Tables? Should we use
+    # separate/parameterised structs?
+    mean::Dict{Symbol, AbstractFloat}
+    std::Dict{Symbol, AbstractFloat}
+
+    """
+        MeanStdScaling() <: Scaling
+
+    Construct a `MeanStdScaling` transform which will populate its `mean` and `std`
+    parameters from the first data it is applied to.
```
glennmoy, line 36:
```suggestion
    parameters from the data it is applied to in a `Dict` indexed by the array indices or 
    Table column names.
```

The idea of "first" to me reads that it applies it to the first column it comes across in the data.

I think you might also want to say that for any `MeanStdScaling` you can only populate it once.

src/scaling.jl
glennmoy: that would be helpful

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+Construct a [`MeanStdScaling`](@ref) using the mean and standard deviation of the data.
+
+# Keyword arguments
+*
+"""
+function MeanStdScaling(data; kwargs...)
+    scaling = MeanStdScaling()
+    populate_stats!(scaling, data; kwargs...)
+    return scaling
+end
```
glennmoy, line 53: I know this method existed in FeatureEngineering but note this is already covered by making `Transform`s themselves callable types. 

So possibly you are overloading this unnecessarily and in a wrong way...

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+    mean::Dict{Symbol, AbstractFloat}
+    std::Dict{Symbol, AbstractFloat}
+
+    """
+        MeanStdScaling() <: Scaling
+
+    Construct a `MeanStdScaling` transform which will populate its `mean` and `std`
+    parameters from the first data it is applied to.
+    """
+    MeanStdScaling() = new(Dict(), Dict())
```
glennmoy, line 38: Would be interested to hear what containers others would suggest to use here...

Arrays would imply that a simple Vector is enough.

But where Tables are concerned you want to make sure you apply the right statistics to the right columns.
So need to store the column names somehow.
Maybe a NamedTuple? But you can't mutate that AFAIK when you instantiate it. Hence the Dict

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+        scaling.std[cname] = std(x)
+    end
+end
+
+"""
+    apply!(A::AbstractArray, ::Transform; dims=:, kwargs...)
+
+Applies [`MeanStdScaling`](@ref) to each element of `A`.
+Optionally specify the `dims` to apply [`MeanStdScaling`](@ref) along certain dimensions,
+and `inverse=true` to reconstruct the originally scaled data from `A`.
```
glennmoy, line 84:
```suggestion
Set `inverse=true` to reconstruct the originally scaled data from `A`.
```

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+For `inverse=false`, all 0 std values get replaced by the value `eps` before scaling.
```
glennmoy, line 85: `eps` is not mentioned in the docstring

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+        x = getproperty(columntable, cname)
+        scaling.mean[cname] = mean(x)
+        scaling.std[cname] = std(x)
+    end
+end
+
+"""
+    apply!(A::AbstractArray, ::Transform; dims=:, kwargs...)
+
+Applies [`MeanStdScaling`](@ref) to each element of `A`.
```
glennmoy, line 82: you also want to explain here (or maybe in the constructor docstring) that
- `apply`  the `MeanStdScaling` the first time populates the statistics and scales the data
- `apply` the `MeanStdScaling` a second time scales the data according to the statistics it contains

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+Optionally specify the `dims` to apply [`MeanStdScaling`](@ref) along certain dimensions,
+and `inverse=true` to reconstruct the originally scaled data from `A`.
+For `inverse=false`, all 0 std values get replaced by the value `eps` before scaling.
+"""
+function apply!(
+    A::AbstractArray, scaling::MeanStdScaling;
+    dims=:, inverse=false, eps=1e-3, kwargs...
+)
+    if isempty(scaling.mean) || isempty(scaling.std)
```
glennmoy, line 91:
```suggestion
    if isempty(scaling.mean) && isempty(scaling.std)
```

Ideally you want them both to be empty.
Have a separate check that errors if only one of them is populated? Since it's ambiguous what to do in that case.

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+
+"""
+    _apply!(
+        A::AbstractArray, t::MeanStdScaling;
+        dims=:, inverse=false, kwargs...
+    )
+
+Applies [`MeanStdScaling`](@ref) to each element of `A`.
+Optionally specify the `dims` to apply [`MeanStdScaling`](@ref) along certain dimensions,
+and `inverse=true` to reconstruct the originally scaled data from `A`.
```
glennmoy, line 113: same comments as above

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+For `inverse=false`, all 0 std values get replaced by the value `eps` before scaling.
+"""
+function _apply!(
+    A::AbstractArray, scaling::MeanStdScaling;
+    name=nothing, inverse=false, eps=1e-3, kwargs...
```
glennmoy, line 118: where does `dims` come into this?

src/scaling.jl
nicoleepp: Can we compute them at construction time instead of at runtime? 

`t = MeanStdScaling(data)` that would calculate the mean and std that would me used anything aply(x, t) the previously computed ones would be used? 

That would enable us to re-use the general apply methods.

nicoleepp:
I noticed you've been merging branches into this one, can you try rebasing instead? It keeps the git history much more clean.

src/scaling.jl
bencottier: > Can we compute them at construction time instead of at runtime?

Good point. I made an additional constructor to do that, but I think it should be the default constructor.

bencottier: > That would enable us to re-use the general apply methods.

The only problem I see with that is passing the `name` keyword argument to `_apply!`, which ensures the slices/columns are scaled in the correct order.

src/scaling.jl
```diff
@@ -0,0 +1,192 @@
...
+        populate_stats!(scaling, table; cols=cols)
+    end
+
+    # Extract a columns iterator that we should be able to use to mutate the data.
+    # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
+    columntable = Tables.columns(table)
+
+    cnames = cols === nothing ? propertynames(columntable) : cols
+    for cname in cnames
+        apply!(getproperty(columntable, cname), scaling; name=cname, kwargs...)
```
bencottier, line 162: @nicoleepp the `name` keyword here and in the array version prevents re-using the general apply methods. It makes sure the correct mean/std is applied to the correct column. Any ideas?

My ideas:
- Require users to keep the order consistent (should be fine for arrays but annoying for tables)
- Add "iter kwargs" to general methods, which get iterated and passed to the inner `_apply!` one-by-one.
- Still have a custom apply method, but it constructs a new MeanStdScaling with the stats in the correct order, then passes that version of the transform to the general method.

bencottier: I would also need to add `kwargs...` to the general apply for tables, in order to pass on `inverse` and `eps` arguments.

src/scaling.jl
bencottier: I forgot to update the docstring

bencottier:
I have done some refactoring to always populate the `mean` and `std` at construction time (either precomputed or computed). I added the `name` keyword argument to the general apply methods for compatibility.

Overall it seems better - much less code in `scaling.jl` now. One thing is `dims` or `cols` has to be specified both at construction and `apply` if used.

src/scaling.jl
bencottier: This caused method ambiguity with `apply!(::AbstractArray, ::Transform; kwargs...)`. I changed this to `_apply!`, which solved the ambiguity but means it goes through the outer `apply` method wastefully.

nicoleepp:
> Note: the notion of dims is consistent with eachslice. Statistics.mean/std and mapslices have the opposite notion of dims (e.g. dims=1 ~ dims=2 and dims=3 ~ dims=[1, 2]). Statistics.mean/std was used in FeatureEngineering. Related issue: #18

Why did you decide to go with dims being consistent with eachslice instead of what we are already doing in FeatureEngineering?

src/scaling.jl
```diff
@@ -0,0 +1,104 @@
...
+abstract type Scaling <: Transform end
+
+"""
+    IdentityScaling
+
+Represents the no-op scaling which simply returns the data it is applied on.
+"""
+struct IdentityScaling <: Scaling end
+# Convenience method
+IdentityScaling(args...; kwargs...) = IdentityScaling()
```
nicoleepp, line 15: I think this is unneccesary?

src/scaling.jl
```diff
@@ -0,0 +1,104 @@
...
+        # Avoid division by 0
+        # If std is 0 then data was uniform, so the scaled value would end up ≈ 0
+        # Therefore the particular `eps` value should not matter much.
+        σ_safe = σ == 0 ? eps : σ
+        A[:] = (A .- μ) ./ σ_safe
+    end
+    return A
+end
+
+function apply(A::AbstractArray, scaling::Scaling; kwargs...)
```
nicoleepp, line 102: Is this meant to override the general apply method? Just wondering why it's needed

test/scaling.jl
```diff
@@ -0,0 +1,393 @@
...
+            expected = zeros(Float64, 3)
+
+            scaling = MeanStdScaling(x)
+
+            @test Transforms.apply(x, scaling) == expected  # default `eps`
+            @test Transforms.apply(x, scaling; eps=1) == expected
+            @test all(isnan.(Transforms.apply(x, scaling; eps=0)))  # 0/0
+        end
+    end
+end
```
nicoleepp, line 393: Does this transform support inds? Would be nice to get a few tests for that as well if it does.

nicoleepp (commented):
Overall looking a lot cleaner now I think

src/scaling.jl
nicoleepp: Passing a name kwarg seems fine to me

src/scaling.jl
nicoleepp: `but means it goes through the outer apply method wastefully.` 

Doesn't the outer method enable the support for dims and inds so that it doesn't need to be re-implemented for your transform?

src/scaling.jl
nicoleepp: > Can the general methods just have the name keyword without affecting other transforms

Seems fine to me

> I would also need to add kwargs... to the general apply! for tables, in order to pass on inverse and eps arguments. I think that should be there anyway.

I agree

bencottier:
> > Note: the notion of dims is consistent with eachslice. Statistics.mean/std and mapslices have the opposite notion of dims (e.g. dims=1 ~ dims=2 and dims=3 ~ dims=[1, 2]). Statistics.mean/std was used in FeatureEngineering. Related issue: #18
> 
> Why did you decide to go with dims being consistent with eachslice instead of what we are already doing in FeatureEngineering?

Because the other transforms so far use dims consistent with eachslice. But I think this is the first transform where `dims` actually changes the results. So given that, and the FeatureEngineering convention, I'm now leaning towards switching over to `mapslices` and its convention. Shall I go ahead with that?

src/scaling.jl
bencottier: We need it to replace FeatureEngineering directly. I'd like it to be unnecessary, but it's used in some tests for GLMModels. Maybe that itself is unnecessary though.

nicoleepp:
> > > Note: the notion of dims is consistent with eachslice. Statistics.mean/std and mapslices have the opposite notion of dims (e.g. dims=1 ~ dims=2 and dims=3 ~ dims=[1, 2]). Statistics.mean/std was used in FeatureEngineering. Related issue: #18
> > 
> > 
> > Why did you decide to go with dims being consistent with eachslice instead of what we are already doing in FeatureEngineering?
> 
> Because the other transforms so far use dims consistent with eachslice. But I think this is the first transform where `dims` actually changes the results. So given that, and the FeatureEngineering convention, I'm now leaning towards switching over to `mapslices` and its convention. Shall I go ahead with that?

The other mutating transforms follow eachslice, but the non mutating follow the mapslices convention. So yeah I think we should move everything to follow the mapslices convention for dims

src/scaling.jl
nicoleepp: Yeah I would say we can update those tests when we switch

src/scaling.jl
bencottier: It's meant to, but it's a cop-out. To use the general apply method we need to deal with the `dims` convention issue.

test/scaling.jl
bencottier: Not currently, but it can support inds. I'm not sure what it would be used for though. If we can get it working with the general `apply` method, then inds should be supported.

bencottier:
Sorry, I made a mess of commits trying to rebase after #19 . I will squash before merging.

src/transformers.jl
```diff
@@ -72,7 +72,11 @@ function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
         end
     end
 
-    return mapslices(x -> _apply(x[inds], t; kwargs...), A, dims=dims)
+    slice_index = 0
+    return mapslices(A, dims=dims) do x
+        slice_index += 1
```
bencottier, line 77: I couldn't find a way to enumerate directly. [Suggestions from discourse](https://discourse.julialang.org/t/is-there-a-way-to-enumerate-mapslices-other-than-i-0-result-mapslices-m/55414) use other packages/functions.

nicoleepp: Does mapslices support multiple dims like dims=[1,2] and if it does, does that cause issues with slice indexing like this? Also if, dims=2 (not the default), does the corresponding std and mean named tuples know that it doesn't apply to the columns anymore?

src/utils.jl
```diff
@@ -11,3 +11,14 @@ function _try_copy(data)
         deepcopy(data)
     end
 end
+
+function invert_dims(A::AbstractArray, dims)
+    ndims(A) == 1 && return dims
+    # TODO: support named dims
```
nicoleepp, line 17: Can you open an issue and link it in the code here?

test/scaling.jl
```diff
@@ -0,0 +1,435 @@
...
+            @testset "dims = 2" begin
+                scaling = MeanStdScaling(A; dims=2)
+                A_1_expected = [0.0 -1.0 1.0; -1.0 0.0 1.0]
+                @test Transforms.apply(A, scaling; dims=2) ≈ A_1_expected atol=1e-5
+            end
+
+            @testset "inds" begin
+                scaling = MeanStdScaling(M)
+                expected = [-0.559017, -1.11803]
+                @test Transforms.apply(M, scaling; inds=[2, 3]) ≈ expected atol=1e-5
```
nicoleepp, line 274: Just for readability, everywhere `expected` is a one-off can we subsitute it for the actual values

```julia
@test Transforms.apply(M, scaling; inds=[2, 3]) ≈ [-0.559017, -1.11803] atol=1e-5
```

bencottier: I think I did this because of line limits, so should I break the statement into multiple lines?

src/transformers.jl
bencottier: > Does mapslices support multiple dims like dims=[1,2] and if it does, does that cause issues with slice indexing like this? 

It does support multiple dims like that. I expect it would cause an issue somewhere, just because we haven't tested it and Murphy's Law. But I don't see what the problem would be - as long as `dims` is consistent between constructing `MeanStdScaling` and applying, the `slice_index` will correspond. `slice_index` is just numbering the slices, whatever those slices are.

> Also if, dims=2 (not the default), does the corresponding std and mean named tuples know that it doesn't apply to the columns anymore?

Not sure I understand right, but again as long as `dims` is the same between constructing and applying, the named tuples will be right. I was thinking about adding `dims` as a field, to check this / avoid specifying `dims` twice, but I want to leave that for now.

src/scaling.jl
```diff
@@ -0,0 +1,89 @@
...
+end
+
+"""
+    MeanStdScaling(data; kwargs...) <: Scaling
+
+Construct a [`MeanStdScaling`](@ref) using the mean and standard deviation of the data.
+
+# Keyword arguments
+* `dims=:`: for `AbstractArray` data, the dimension(s) to compute statistics along.
+* `cols=nothing`: for `Table` data, the column names to compute statistics for.
```
nicoleepp, line 40: Can you add a warning or note that dims and cols just be specified the same way when constructing and applying or else the resulsts will be inconsistent?

src/transformers.jl
nicoleepp: Ok I think this is fine then

src/transformers.jl
```diff
@@ -106,8 +114,9 @@ Optionally specify the `dims` to apply the [`Transform`](@ref) along certain dim
 function apply!(A::AbstractArray, t::Transform; dims=:, kwargs...)
     dims == Colon() && return _apply!(A, t; kwargs...)
 
-    for x in eachslice(A; dims=dims)
-        _apply!(x, t; kwargs...)
+    _dims = invert_dims(A, dims)  # opposite convention to `eachslice`
```
nicoleepp, line 117: Can you add a blurb on what direction dims and inds correspond to to this docstring?

src/utils.jl
nicoleepp: Let me know if this changes after Rory's input or not.

test/scaling.jl
nicoleepp: The BlueStyle/Invenia Julia line limits are recommendations more than hard rules. I think in this case for readability is one where having it go over the line length on one line is more valuable than adhering to line length

src/utils.jl
bencottier: No change - #20

nicoleepp (approved):

PR #19, opened 2021-02-17
Add OneHotEncoding Transform

### LINKED ISSUE #7: One hot encoding transform
A one-hot-encoding transform for categorical variables

MWE
```julia
x = ["foo", "bar", "baz"]

ohe = OneHotEncoding()

Transform.apply(x, ohe)

# output
3×3 Array{Int64,2}:
 1  0  0
 0  1  0
 0  0  1
```

nicoleepp:
Closes #7

### DIFF
--- a/src/one_hot_encoding.jl
+++ b/src/one_hot_encoding.jl
@@ -1,0 +1,41 @@
+"""
+    OneHotEncoding <: Transform
+
+One-hot encode the categorical value for each target element.
+
+Construct a n-by-p binary matrix, given a `Vector` of target data `x` (of length n) and a
+`Vector` of all unique possible values in x (of length p).
+
+The element [i, j] is `1` if the i^th target in `x` corresponds to the j^th possible
+value and `0` otherwise.
+
+Note that this Transform does not support specifying dims other than `:` (all dims) because
+it is a one-to-many transform (for example a `Vector` input produces a `Matrix` output).
+"""
+struct OneHotEncoding <: Transform
+    categories::Dict{Any, Int}
+
+    function OneHotEncoding(possible_values::AbstractVector)
+        if length(unique(possible_values)) < length(possible_values)
+            throw(ArgumentError("Expected an ordered list of all unique possible values"))
+        end
+
+        # Create a dictionary that maps unique values in the input array to column positions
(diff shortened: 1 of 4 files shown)


src/one_hot_encoding.jl
```diff
@@ -0,0 +1,41 @@
...
+
+Note that this Transform does not support specifying dims other than `:` (all dims) because
+it is a one-to-many transform (for example a `Vector` input produces a `Matrix` output).
+"""
+struct OneHotEncoding <: Transform
+    categories::Dict{Any, Int}
+
+    function OneHotEncoding(possible_values::AbstractVector)
+        if length(unique(possible_values)) < length(possible_values)
+            throw(ArgumentError("Expected an ordered list of all unique possible values"))
```
nicoleepp, line 20: Should test this

bencottier, line 20: Do the elements need to be a consistent order? Or does "ordered list" just refer to the type?

test/one_hot_encoding.jl
```diff
@@ -0,0 +1,118 @@
+@testset "one hot encoding" begin
```
bencottier, line 1: As well as testing non-unique `categories` (as you noted), could we test
- data having a value not in `categories`
- case where `length(categories) > length(x)` - both (a) to test non-square matrix size both ways, and (b) to test `x` not having all the categories

bencottier (commented):
I think we should have a couple more kinds of tests, it looks good otherwise.

src/one_hot_encoding.jl
nicoleepp: They just need to be in the order of expected column output. so for input of 24 hours they should be ordered from Hour(1) to Hour(24). Then if it is applied on x = [Hour (2), Hour(2), Hour(1)] the result would be

```
 0  1  0  0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0  1  0  0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 1  0  0  0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 ```

bencottier (approved):

PR #23, opened 2021-02-19
Set main as devbranch so docs should build

### LINKED ISSUE #22: Docs are not building.
I suspect the documentor key needs to be setup

oxinabox:
Hopefully closes #22 
if so then it was not a missing documentor key

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,4 +20,5 @@ makedocs(;
 
 deploydocs(;
     repo="github.com/invenia/Transforms.jl",
+    devbranch = "main",
 )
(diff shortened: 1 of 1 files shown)


docs/make.jl
```diff
@@ -20,4 +20,5 @@ makedocs(;
 
 deploydocs(;
     repo="github.com/invenia/Transforms.jl",
+    devbranch = "main",
```
oxinabox, line 23:
```suggestion
    devbranch = "main",
    push_preview = "true",
```

docs/make.jl
```diff
@@ -20,4 +20,6 @@ makedocs(;
...
+    push_preview = "true",
```
oxinabox, line 24:
```suggestion
    push_preview = true,
```

oxinabox:
woohoo it works

nicoleepp (approved):

PR #24, opened 2021-02-19
Add IdentityScaling

nicoleepp:
I think there was a mixup in review, this is a follow up to #17

### DIFF
--- a/src/scaling.jl
+++ b/src/scaling.jl
@@ -1,9 +1,20 @@
 """
-    Scaling <: Transform
+    AbstractScaling <: Transform
 
 Linearly scale the data as `ax + b`, according to some statistics `a` and `b`.
 """
-abstract type Scaling <: Transform end
+abstract type AbstractScaling <: Transform end
+
+"""
+    IdentityScaling
+
+Represents the no-op scaling which simply returns the `data` it is applied on.
+"""
+struct IdentityScaling <: AbstractScaling end
+
+_apply(x, scaling::IdentityScaling; kwargs...) = x
+_apply!(x, scaling::IdentityScaling; kwargs...) = _apply(x, scaling; kwargs...)
+
 
 """
     MeanStdScaling(mean, std) <: Scaling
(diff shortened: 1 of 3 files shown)


src/scaling.jl
```diff
@@ -1,9 +1,20 @@
...
+Represents the no-op scaling which simply returns the `data` it is applied on.
+"""
+struct IdentityScaling <: AbstractScaling end
+
+_apply(x, scaling::IdentityScaling; kwargs...) = x
+_apply!(x, scaling::IdentityScaling; kwargs...) = _apply(x, scaling; kwargs...)
+
 
 """
     MeanStdScaling(mean, std) <: Scaling
```
glennmoy, line 20:
```suggestion
    MeanStdScaling(mean, std) <: AbstractScaling
```

src/scaling.jl
```diff
@@ -1,9 +1,20 @@
...
-    Scaling <: Transform
+    AbstractScaling <: Transform
 
 Linearly scale the data as `ax + b`, according to some statistics `a` and `b`.
 """
-abstract type Scaling <: Transform end
+abstract type AbstractScaling <: Transform end
+
+"""
+    IdentityScaling
```
glennmoy, line 9:
```suggestion
    IdentityScaling <: AbstractScaling
```

test/scaling.jl
```diff
@@ -1,4 +1,173 @@
 @testset "scaling" begin
+
+    @testset "IdentityScaling" begin
+        scaling = IdentityScaling()
+        @test scaling isa Transform
+
+        @testset "Vector" begin
+            x = [1., 2., 3.]
+            expected = [1., 2., 3.]
```
glennmoy, line 9: minor nit-pick, but is it worth specifying `expected` in the case of `IdentityScaling`? since the operation is idempotent would it not be easier/clearer to say `@test Transforms.apply(x, scaling) == x` for all these tests?

glennmoy (approved):

test/scaling.jl
nicoleepp: I like using expected as it ensures `x` hasn't changed

PR #27, opened 2021-02-23
WIP: Use mapslices in LinearCombination

glennmoy:
Follow up to #25 

LinearCombination still uses eachslices, and thus still uses the opposite convention to `dims` compared to our other Transforms.

Before - no distinction between application to rows vs cols, default is `dims=1`
```julia

julia> lc = LinearCombination([1, 2])
LinearCombination(Real[1, 2])

julia> Transforms.apply(A, lc; dims=1, inds=[1, 2])
3-element Array{Int64,1}:
  5
 14
 23

julia> Transforms.apply(A, lc; dims=2, inds=[1, 2])
3-element Array{Int64,1}:
  9
 12
 15
```

Now - applying to rows vs cols reduces the array accordingly, default is now `dims=2`
```julia

julia> lc = LinearCombination([1, 2])
LinearCombination(Real[1, 2])

julia> Transforms.apply(A, lc; dims=1, inds=[1, 2])
1×3 Array{Int64,2}:
 9  12  15

julia> Transforms.apply(A, lc; dims=2, inds=[1, 2])
3×1 Array{Int64,2}:
  5
 14
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Transforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,9 +1,11 @@
 name = "Transforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.1.1"
+version = "0.1.0"
 
 [deps]
+AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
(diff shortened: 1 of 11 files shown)


Project.toml
```diff
@@ -4,6 +4,8 @@ authors = ["Invenia Technical Computing Corporation"]
 version = "0.1.0"
 
 [deps]
+AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
glennmoy, line 8: remove these

src/linear_combination.jl
```diff
@@ -35,25 +35,23 @@ function apply(x::AbstractVector, LC::LinearCombination; inds=Colon())
 end
 
 """
-    apply(x::AbstractMatrix, LC::LinearCombination; dims=1, inds=Colon())
+    apply(A::AbstractMatrix, LC::LinearCombination; dims=2, inds=Colon())
```
glennmoy, line 38: changed `x` to `A` more on a whim but also so I could use `x` in the `mapslices` below

test/runtests.jl
```diff
@@ -7,6 +7,9 @@ using Transforms: _try_copy, _periodic
 using Test
 using TimeZones
 
+# TODO - this is needed to make mapslices work with AxisArrays - should be removed.
+AxisArrays.reduced_indices(X::Tuple{Vararg{Axis}}, r::UnitRange) = AxisArrays.reduced_indices(X, Tuple(collect(r)))
```
glennmoy, line 11: I hate this. I think this should just be added to AxisArrays but depending how difficult that is to get in it might just be easier to ditch AxisArrays altoghther.

glennmoy: https://github.com/JuliaArrays/AxisArrays.jl/pull/194

PR #28, opened 2021-02-23
Allow single column name for `cols`

### LINKED ISSUE #26: Have `apply` return single column array when `cols` is a single column name
The non-mutating `apply` in transformers.jl [always returns a vector](https://github.com/invenia/Transforms.jl/blob/83c8558aee22dbd343eceab659d132a80897c5fd/src/transformers.jl#L99) of the transformed columns specified in `cols`. `cols` is assumed to be a collection.

For convenience, we could add that if `cols` is a single column name we return the single column vector, rather than wrapping it in a 1-element vector.

I vote that we do the above using a bare String or Symbol (i.e. **not** in a collection), and still allow e.g. `cols=[:colname]` to return the 1-element vector. MWE of what this means:

```
julia> p = Power(3);

julia> df = DataFrame(:a => [1, 2, 3], :b => [4, 5, 6], :c => [7, 8, 9]);

julia> Transforms.apply(df, p; cols=[:a, :c])
2-element Array{Array{Int64,1},1}:
 [1, 8, 27]
 [343, 512, 729]

julia> Transforms.apply(df, p; cols=[:a])
 [1, 8, 27]

julia> Transforms.apply(df, p; cols=:a)
3-element Array{Int64,1}:
  1
  8
 27
```

bencottier:
Closes #26 

This is the behaviour using a single column name (i.e. not a list):
- `apply` returns the single transformed column, "unwrapped"
  - Except for `LinearCombination`, because it always produces one column anyway
- `apply!` allows it, but gives the same result

### DIFF
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -64,9 +64,7 @@ Applies the [`LinearCombination`](@ref) to each of the specified cols in `x`.
 If no `cols` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
 """
 function apply(x, LC::LinearCombination; cols=nothing)
-    if cols isa Union{Symbol, AbstractString}  # want same behaviour for single column
-        cols = [cols]
-    end
+    cols = _to_vec(cols)  # handle single column name
 
     # Error if dimensions don't match
     num_cols = cols === nothing ? length(Tables.columnnames(x)) : length(cols)
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -64,6 +64,10 @@ Applies the [`LinearCombination`](@ref) to each of the specified cols in `x`.
 If no `cols` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
 """
 function apply(x, LC::LinearCombination; cols=nothing)
+    if cols isa Union{Symbol, AbstractString}  # want same behaviour for single column
+        cols = [cols]
+    end
+
     # Error if dimensions don't match
     num_cols = cols === nothing ? length(Tables.columnnames(x)) : length(cols)
     _check_dimensions_match(LC, num_cols)
--- a/src/transformers.jl
+++ b/src/transformers.jl
@@ -110,10 +110,14 @@ function apply(table, t::Transform; cols=nothing, kwargs...)
 
     cnames = cols === nothing ? propertynames(columntable) : cols
 
-    return [
-        _apply(getproperty(columntable, cname), t; name=cname, kwargs...)
-        for cname in cnames
-    ]
+    if cnames isa Union{Symbol, AbstractString}  # want unwrapped single column
+        return _apply(getproperty(columntable, cnames), t; name=cnames, kwargs...)
+    else
+        return [
+            _apply(getproperty(columntable, cname), t; name=cname, kwargs...)
+            for cname in cnames
+        ]
+    end
 end
 
 """
@@ -125,6 +129,10 @@ If no `cols` are specified, then the [`Transform`](@ref) is applied to all colum
 function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
     # TODO: We could probably handle iterators of tables here
     Tables.istable(table) || throw(MethodError(apply!, (table, t)))
+
+    if cols isa Union{Symbol, AbstractString}  # want same behaviour for single column
+        cols = [cols]
+    end
 
     # Extract a columns iterator that we should be able to use to mutate the data.
     # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
--- a/src/transformers.jl
+++ b/src/transformers.jl
@@ -110,13 +110,13 @@ function apply(table, t::Transform; cols=nothing, kwargs...)
 
     cnames = cols === nothing ? propertynames(columntable) : cols
 
-    if cnames isa Union{Symbol, AbstractString}  # want unwrapped single column
-        return _apply(getproperty(columntable, cnames), t; name=cnames, kwargs...)
-    else
+    if cnames isa Union{Tuple, AbstractArray}
         return [
             _apply(getproperty(columntable, cname), t; name=cname, kwargs...)
             for cname in cnames
         ]
+    else  # return unwrapped single column
+        return _apply(getproperty(columntable, cnames), t; name=cnames, kwargs...)
     end
 end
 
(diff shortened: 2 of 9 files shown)


src/linear_combination.jl
```diff
@@ -64,6 +64,10 @@ Applies the [`LinearCombination`](@ref) to each of the specified cols in `x`.
 If no `cols` are specified, then the [`LinearCombination`](@ref) is applied to all columns.
 """
 function apply(x, LC::LinearCombination; cols=nothing)
+    if cols isa Union{Symbol, AbstractString}  # want same behaviour for single column
```
nicoleepp, line 67: We can use multiple dispatch here instead wich is generally more performant

```julia
function apply(x, LC::LinearCombination; cols::Union{Symbol, AbstractString})
     apply(x, LC; cols=[cols]
 end
 ```

File: test/linear_combination.jl [hunk 2]
```diff
@@ -218,5 +226,13 @@
             @test Transforms.apply(df, lc; cols=cols) == expected
             @test lc(df; cols=cols) == expected
         end
+
+        @testset "single col" begin
+            lc_single = LinearCombination([-1])
+
+            @test Transforms.apply(df, lc_single; cols=:a) == [-1, -2, -3]
```
nicoleepp, line 233: Does this actually work for the string `cols="a"`? We are checking if the type is AbstractString but don't have tests for it

test/one_hot_encoding.jl
```diff
@@ -121,6 +122,7 @@
         @test Transforms.apply(df, ohe) == expected
 
         @test Transforms.apply(df, ohe; cols=[:a]) == [[1 0 0 0 0; 0 1 0 0 0]]
```
nicoleepp, line 124: Wouldn't we also want this to change to not return a nested matrix?

nicoleepp: I thought this is what you were gonna change from your post on slack so i probably misunderstood but I would vote we don't wrap this for single cols (even if they get passed in as an array)

bencottier: My reasoning is there should still be the option to return it wrapped, and the `[:a]` vs. `:a` syntax is intuitive for that (thought come to think of it, that should be documented). Returning it wrapped keeps it consistent, e.g. if the user has some generalised code that doesn't know how many columns will be given.

glennmoy: >Returning it wrapped keeps it consistent

I guess that makes sense. One thing in your favour is how extracting columns from a dataframe works.
`:a` vs `[:a]` does not return the same thing.

```julia
julia> df = DataFrame(:a=>[1, 2, 3, 4, 5], :b=>[5, 4, 3, 2, 1], :c=>[0, 1, 0, 1, 0]);

julia> df[:, :a]
5-element Array{Int64,1}:
 1
 2
 3
 4
 5

julia> df[:, [:a]]
5×1 DataFrame
│ Row │ a     │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 1     │
│ 2   │ 2     │
│ 3   │ 3     │
│ 4   │ 4     │
│ 5   │ 5     │
```

I wonder are there other examples we could lean on to inform this decision?

bencottier: Another example (note my confirmation bias):

```
julia> x = [4, 5, 6];

julia> x[1]
4

julia> x[[1]]
1-element Array{Int64,1}:
 4
```

bencottier: Example the other way:

```
julia> df = DataFrame(:a => [1, 2, 3], :b => [4, 5, 6]);

julia> combine(df, :a => sum)
1×1 DataFrame
│ Row │ a_sum │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 6     │

julia> combine(df, [:a] => sum)
1×1 DataFrame
│ Row │ a_sum │
│     │ Int64 │
├─────┼───────┤
│ 1   │ 6     │
```

nicoleepp: To keep it simple and consitent I now think we should now do the :a => [...] and [:a] => [[...]] (as this PR already does)

I would argue the last example that Ben posted is actually wrapping both of them (a third option that we for sure don't want I imagine).

src/linear_combination.jl
bencottier: > So you'll need something like a _to_vec method

Just note I'm trying to allow `cols` as a `Tuple`, so I want to dispatch the opposite way on `Union{Symbol, AbstractString}` rather than `AbstractArray`.

bencottier: Maybe I should just do `Union{Tuple, AbstractArray}`

> > So you'll need something like a _to_vec method
> 
> Just note I'm trying to allow `cols` as any collection including `Tuple`, so I want to dispatch the opposite way on `Union{Symbol, AbstractString}` rather than `AbstractArray`.

Maybe just `Union{Tuple, AbstractArray}` is good enough and clearer, actually.

test/linear_combination.jl
bencottier: I've refactored so it doesn't check `AbstractString` now, but the reason I did that is DataFrames supports indexing columns as strings.

src/transformers.jl
```diff
@@ -138,6 +142,8 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
     # TODO: We could probably handle iterators of tables here
     Tables.istable(table) || throw(MethodError(apply!, (table, t)))
 
+    cols = _to_vec(cols)
```
bencottier, line 145:
```suggestion
    cols = _to_vec(cols)  # handle single column name
```

src/utils.jl
```diff
@@ -22,3 +22,8 @@ function invert_dims(A::AbstractArray, dims)
 
     return inverted_dims
 end
+
+_to_vec(x::AbstractArray) = x
+_to_vec(x::Tuple) = x
+_to_vec(x::Nothing) = x
+_to_vec(x) = [x]
```
oxinabox, line 29: You may want to cross-reference `FiniteDifferences.to_vec`.
Which if you need that behavour we could split into its own package

nicoleepp (approved):

src/utils.jl
bencottier: Thanks, definitely something to keep in mind. I think this is a simple and special case that should live on its own for now. In fact calling it `_to_vec` may just be confusing. Based on its current use a better name might be `_convert_cols`.

src/transformers.jl
```diff
@@ -96,10 +96,14 @@ function apply(table, t::Transform; cols=nothing, kwargs...)
 
     cnames = cols === nothing ? propertynames(columntable) : cols
 
-    return [
-        _apply(getproperty(columntable, cname), t; name=cname, kwargs...)
-        for cname in cnames
-    ]
+    if cnames isa Union{Tuple, AbstractArray}
```
glennmoy, line 99: can we not use `_to_vec` here as well?

glennmoy (approved):

src/transformers.jl
bencottier: Hmm, I think it's different to the `_to_vec` for cols.

Here: If the value is a multi-element array, return that. Else (if it's a one-element array), return the element.

Whereas for cols: if the value is _any_ array, return that. Else (if it's not an array), return it as an array.

bencottier: Incidentally, this reminds me to update the docstring with :a vs [:a] behaviour.

PR #31, opened 2021-02-24
Use Bool by default for OneHotEncoding Transform

### LINKED ISSUE #7: One hot encoding transform

nicoleepp:
Closes #7

### DIFF
--- a/src/one_hot_encoding.jl
+++ b/src/one_hot_encoding.jl
@@ -6,16 +6,16 @@ One-hot encode the categorical value for each target element.
 Construct a n-by-p binary matrix, given a `Vector` of target data `x` (of length n) and a
 `Vector` of all unique possible values in x (of length p).
 
-The element [i, j] is `1` if the i^th target in `x` corresponds to the j^th possible
-value and `0` otherwise.
+The element [i, j] is `true` if the i^th target in `x` corresponds to the j^th possible
+value and `false` otherwise.
 
 Note that this Transform does not support specifying dims other than `:` (all dims) because
 it is a one-to-many transform (for example a `Vector` input produces a `Matrix` output).
 """
-struct OneHotEncoding <: Transform
-    categories::Dict{Any, Int}
+struct OneHotEncoding{T} <: Transform
+    categories::Dict{T, Int}
 
-    function OneHotEncoding(possible_values::AbstractVector)
+    function OneHotEncoding(possible_values::AbstractVector{T}) where T
         if length(unique(possible_values)) < length(possible_values)
             throw(ArgumentError("Expected a list of all unique possible values"))
         end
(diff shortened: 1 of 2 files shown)


src/one_hot_encoding.jl
```diff
@@ -6,35 +6,34 @@ One-hot encode the categorical value for each target element.
...
         # in the sparse matrix that results from applying the OneHotEncoding transform
         categories = Dict(value => i for (i, value) in enumerate(possible_values))
-        return new(categories)
+        return new{T}(categories)
     end
 end
 
 function _apply(x, encoding::OneHotEncoding; kwargs...)
     n_categories = length(encoding.categories)
+    results = zeros(Bool, length(x), n_categories)
```
rofinn, line 32: If we parameterized the `encoding` type by `R<:Real` then we could make this more flexible as `zeros(R, ...)`. The `true` below would just get converted on `setindex!`.

rofinn: Thanks, LGTM 👍🏻

src/one_hot_encoding.jl
```diff
@@ -1,40 +1,44 @@
...
+function OneHotEncoding(possible_values::AbstractVector{T}) where T
+    return OneHotEncoding{Bool}(possible_values)
+end
 
-    results = zeros(Int, length(x), n_categories)
+function _apply(x, encoding::OneHotEncoding{R}; kwargs...) where R <: Real
+    n_categories = length(encoding.categories)
+    results = zeros(R, length(x), n_categories)
 
-    @views for (i, value) in enumerate(x)
```
bencottier, line 35: Did you remove `@views` just because it seems unnecessary? I don't understand it well myself.

bencottier (approved):
Looks good to me. I think just edit the title e.g. "Use Bool by default for OneHotEncoding Transform".

src/one_hot_encoding.jl
```diff
@@ -1,40 +1,44 @@
...
+The element [i, j] is `true` if the i^th target in `x` corresponds to the j^th possible
+value and `false` otherwise. Note that `R`can be specified to determine the return type
+of results. It defaults to a `Matrix` of `Bool`s.
 
 Note that this Transform does not support specifying dims other than `:` (all dims) because
 it is a one-to-many transform (for example a `Vector` input produces a `Matrix` output).
 """
-struct OneHotEncoding <: Transform
-    categories::Dict{Any, Int}
+struct OneHotEncoding{R<:Real, T} <: Transform
```
glennmoy, line 16: do we need to parameterise this by `T` as well? It doesn't seem to be used for anything - and we don't mention it in the docstring.

glennmoy (approved):

src/one_hot_encoding.jl
nicoleepp: It's used for space efficiency so we don't need to store a Dict of `Any`

PR #32, opened 2021-02-24
Relax period and phase_shift type restrictions

### LINKED ISSUE #29: Do not restrict `period` and `phase_shift` to the same concrete type
[This](https://github.com/invenia/Transforms.jl/blob/83c8558aee22dbd343eceab659d132a80897c5fd/src/periodic.jl#L21-L22) restricts them to the same concrete type.

```
julia> Periodic(sin, Week(1), Day(5))
ERROR: MethodError: no method matching Periodic(::typeof(sin), ::Week, ::Day)
Closest candidates are:
  Periodic(::Any, ::T) where T at /Users/bencottier/.julia/packages/Transforms/NdL6j/src/periodic.jl:36
  Periodic(::Any, ::T, ::T) where T at /Users/bencottier/.julia/packages/Transforms/NdL6j/src/periodic.jl:24
Stacktrace:
 [1] top-level scope at REPL[99]:1
```

It was unintended, only meant to restrict to the same abstract type, i.e. both `Real` or both `Period`.

We need to relax it to exactly match FeatureEngineering, but I don't think it affects results, so I'm not sure if it's worth it.

bencottier:
Closes #29 

A couple of tweaks along the way:
- Change `ArgumentError` to `DomainError` for non-positive period, and test this for `TimeType` as well as `Real`
- Refactor other methods so they don't confuse the parametric type variable `P`

### DIFF
--- a/src/periodic.jl
+++ b/src/periodic.jl
@@ -1,7 +1,10 @@
 """
-    Periodic{T}(f, period::T, phase_shift::T) <: Transform
+    Periodic{P, S}(f, period::P, phase_shift::S) <: Transform
 
 Applies a periodic function `f` with provided `period` and `phase_shift` to the data.
+
+`period` and `phase_shift` must have the same supertype of `Real` or `Period`, depending on
+whether the data is `Real` or `TimeType` respectively.
 
 !!! note
     For `TimeType` data, the result will change depending on the type of `period` given,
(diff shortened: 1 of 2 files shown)


src/periodic.jl
```diff
@@ -1,8 +1,11 @@
 """
-    Periodic{T}(f, period::T, phase_shift::T) <: Transform
+    Periodic{P, S}(f, period::P, phase_shift::S) <: Transform
 
 Applies a periodic function `f` with provided `period` and `phase_shift` to the data.
 
+`period` and `phase_shift` must have the same supertype of `Real` or `Period`, depending on
```
glennmoy, line 6:
```suggestion
The `period` and `phase_shift` must have the same supertype of `Real` or `Period`, depending on
```

src/periodic.jl
```diff
@@ -1,8 +1,11 @@
 """
-    Periodic{T}(f, period::T, phase_shift::T) <: Transform
+    Periodic{P, S}(f, period::P, phase_shift::S) <: Transform
```
glennmoy, line 2:
```suggestion
    Periodic{P, S}(f, period::P, [phase_shift::S]) <: Transform
```

if `phase_shift` is optional this communicates it in the function signature, but it may not be necessary if you have a separate docstring for the other method.

glennmoy (approved):

src/periodic.jl
```diff
@@ -12,18 +15,21 @@ Applies a periodic function `f` with provided `period` and `phase_shift` to the
...
-    phase_shift::T
+    period::P
+    phase_shift::S
 
-    function Periodic(f, period::T, phase_shift::T) where T
-        period > zero(T) || throw(ArgumentError("period must be strictly positive."))
-        return new{T}(f, period, phase_shift)
+    function Periodic(f, period::P, phase_shift::S) where {P, S}
+        if !((P <: Real && S <: Real) || (P <: Period && S <: Period))
+            throw(ArgumentError("period and phase_shift must have the same supertype"))
```
glennmoy, line 29: I guess this gives a more informative error message compared to my suggestion which would have just been a MethodError.

PR #37, opened 2021-03-01
Add initial docs

### LINKED ISSUE #36: Add initial docs

bencottier:
Closes #36 

Includes pages for an introduction, the API, overview of how `Transform`s work, and a longer example.

Note the built docs can be previewed by clicking the "details" link next to the documenter/deploy check.


docs/src/transforms.md
```diff
@@ -0,0 +1,96 @@
...
+ 9.0
+```
+
+## Applying a `Transform`
+
+There are three main ways to apply a `Transform` - suppose it is called `t`:
+
+* `Transforms.apply(data, t; kwargs...)` is non-mutating, returning the transformed data without modifying the original data.
+* `t(data; kwargs...)` is equivalent to `Transforms.apply(data, t; kwargs...)`
+* `Transforms.apply!(data, t; kwargs...)` is mutating, returning the modified `data` with the same type.
```
glennmoy, line 37: Instead of laying this out in a way akin to a docstring, why not demonstrate it by expanding on the example above?

docs/src/transforms.md
```diff
@@ -0,0 +1,96 @@
...
+
+A single `Transform` can be applied to different data types and in different ways. The two main data types supported are `AbstractArray`s and [`Table`s](https://github.com/JuliaData/Tables.jl).
+
+### `AbstractArray` data
+
+For `AbstractArray` data, some `Transform`s support a `dims` keyword argument in their `apply` methods. This will apply the `Transform` to slices of the array along dimensions determined by `dims`. For example, given a `Matrix`, `dims=1` applies to each column, and `dims=2` applies
+to each row. This convention is similar to `Statistics.mean(M; dims=2)` returning the mean of each row in matrix `M`.
```
glennmoy, line 44: I feel like we need to tell 3 different things here and we may want to split them up a bit better

1. How a vanilla `Transform` is applied to an array (e.g. Power)
2. How `inds` works so one can only apply to certain slices 
3. How `dims` works - demonstrated by a slightly more complex transform (e.g. `LinearCombination`).

> This convention is similar to `Statistics.mean(M; dims=2)` returning the mean of each row in matrix `M`.

I think this is helpful to mention (as it will immediately clarify the output to users) but we should use it as a complement to explaining how we use `mapslices` under the hood.

docs/src/transforms.md
```diff
@@ -0,0 +1,96 @@
...
+julia> df = DataFrame(
+           :time => DateTime(2021, 2, 27, 12):Hour(1):DateTime(2021, 2, 27, 14),
+           :temperature_A => [18.1, 19.5, 21.1],
+           :temperature_B => [16.2, 17.2, 17.5],
+       );
+
+julia> feature_df = DataFrame(
+           :hour_of_day => FeatureTransforms.apply(df, HoD(); cols=:time),
+           :aggregate_temperature => FeatureTransforms.apply(df, LinearCombination([0.5, 0.5]); cols=[:temperature_A, :temperature_B])
+       )
```
glennmoy, line 84: Hmm...I'm just wondering if we should tidy this up a bit

```julia
lc = LinearCombination([0.5, 0.5])

df[:, :hour_of_day] = FeatureTransforms.apply(df, HoD(); cols=:time)
df[:, :mean_temp] = FeatureTransforms.apply(df, lc ; cols=[:temperature_A, :temperature_B])
```

docs/src/examples.md
```diff
@@ -0,0 +1,148 @@
...
+│ 18  │ 2018-09-10T17:00:00 │ 21.15       │ 56.8442  │ -0.965926       │
+│ 19  │ 2018-09-10T18:00:00 │ 19.95       │ 60.2864  │ -1.0            │
+│ 20  │ 2018-09-10T19:00:00 │ 18.35       │ 64.7786  │ -0.965926       │
+│ 21  │ 2018-09-10T20:00:00 │ 16.65       │ 70.7941  │ -0.866025       │
+│ 22  │ 2018-09-10T21:00:00 │ 14.95       │ 77.3468  │ -0.707107       │
+│ 23  │ 2018-09-10T22:00:00 │ 13.85       │ 83.0855  │ -0.5            │
+│ 24  │ 2018-09-10T23:00:00 │ 12.75       │ 86.9574  │ -0.258819       │
+```
+
+Now suppose we want to use the first 22 hours as training data and the last 2 hours as test data. Our input features are the temperature, humidity, and periodic encodings for the current hour, and the outputs to predict are the temperature and humidity for the next hour.
```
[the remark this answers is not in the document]
bencottier: Do you think the smaller examples in `transforms.md` are enough? If not I'm wondering whether to find a way to incorporate those here, or make a second example. Any ideas?

docs/src/transforms.md
bencottier: Yeah, clear headed sections for each thing would be kinder to the reader.

> I think this is helpful to mention (as it will immediately clarify the output to users) but we should use it as a complement to explaining how we use mapslices under the hood.

Not sure what you mean - you want to explain in the doc how we use `mapslices` under the hood?

glennmoy: >Not sure what you mean - you want to explain in the doc how we use mapslices under the hood?

As in, we should say something like 

> The `dims` argument uses the convention used by `mapslices`, which is called behind the scenes when applying transformations to slices of data. In practice, this means that users can expect the `dims` keyword to behave exactly as `mean(A; dims=d)` would; the transformation will be applied to the elements along the dimension `d` and, for operations like `mean` or `sum`, reduce across this dimension.

morris25 (commented):
LGTM

docs/src/transforms.md
```diff
@@ -0,0 +1,246 @@
...
+    So `MeanStdScaling` should only be applied to the same data type and for the same dimensions or subset of column names specified in construction.
+
+## Applying to `AbstractArray`
+
+### Default
+
+Without specifying optional arguments, a `Transform` is applied to every element of an `AbstractArray` and in an element-wise fashion:
+
+```jldoctest transforms
+julia> M = [2 4; 1 5; 3 6];
```
morris25, line 91: You might want to change this to a float array to avoid failing the 32 bit tests

docs/src/index.md
```diff
@@ -1,12 +1,17 @@
...
-```@index
-```
+FeatureTransforms.jl provides utilities for performing feature engineering in machine learning pipelines.
+FeatureTransforms supports operations on `AbstractArray`s and [`Table`](https://github.com/JuliaData/Tables.jl)s.
+
+There are three key parts of the Transforms.jl API:
+
+* A [`Transform`](@ref about-transforms) defines a transformation of data, for example normalisation or a periodic function.
+* An `apply` method applies a [`Transform`](@ref about-transforms) to data, in a manner determined by the data type and specified dimensions, column names, indices, and other `Transform`-specific parameters.
+* The `transform` interface can be used to define a feature engineering pipeline, which comprises a collection of [`Transform`](@ref about-transforms)s to be peformed on some data.
```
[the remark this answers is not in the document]
glennmoy: Yeah, maybe just mention the method is there to be overloaded when Transforms are used in feature engineering pipelines. 

We make no restrictions on it now but when we do it'll be a breaking change.

docs/src/transforms.md
```diff
@@ -10,77 +11,199 @@ DocTestSetup = quote
...
+ 9.0
+```
+
+A single `Transform` instance can be applied to different data types, with support for `AbstractArray`s and [`Table`s](https://github.com/JuliaData/Tables.jl).
+
+!!! note
+
+  Some `Transform` subtypes have restrictions on how they can be applied once constructed.
+  For instance, `MeanStdScaling` stores the mean and standard deviation of some data for specified dimensions or column names.
+  So `MeanStdScaling` should only be applied to the same data type and for the same dimensions or subset of column names specified in construction.
```
glennmoy: leave this in for now and I'll address it in #42

docs/src/transforms.md
```diff
@@ -10,77 +11,199 @@ DocTestSetup = quote
...
+julia> FeatureTransforms.apply(M, p)
+3×2 Array{Int64,2}:
+ 4  16
+ 1  25
+ 9  36
+```
+
+### Applying to specific array indices with `inds`
+
+The non-mutating `apply` method has an `inds` keyword argument to apply the `Transform` to certain indices of an array.
```
glennmoy, line 104: so does `apply!` right? it just passes its kwargs to `apply`

docs/src/transforms.md
```diff
@@ -10,77 +11,199 @@ DocTestSetup = quote
...
+  The non-mutating `apply` method for `Table` data returns a `Vector` of `Vector`s, one for each column, e.g.
+
+  ```julia-repl
+  julia> FeatureTransforms.apply(nt, scaling)
+  2-element Array{Array{Float64,1},1}:
+   [-2.0, -3.0, -1.0]
+   [-6.0, -5.0, -4.0]
+  ```
+
+  This is so users are free to decide what to name the results of the transformation, whether to append to the original table, etc.
```
glennmoy: Would be good to put this directly under the sentence above. I couldn't remember why we did this but knew there was a reason.

docs/src/transforms.md
```diff
@@ -10,77 +11,199 @@ DocTestSetup = quote
...
-           :temperature_B => [16.2, 17.2, 17.5],
-       );
+julia> hod = HoD()
+
+julia> lc = LinearCombination([0.5, 0.5])
 
 julia> feature_df = DataFrame(
-           :hour_of_day => FeatureTransforms.apply(df, HoD(); cols=:time),
-           :aggregate_temperature => FeatureTransforms.apply(df, LinearCombination([0.5, 0.5]); cols=[:temperature_A, :temperature_B])
+           :hour_of_day => hod(df; cols=:time),
```
glennmoy, line 211: where is `df` defined now?

docs/src/examples.md
```diff
@@ -0,0 +1,148 @@
...
+│ 23  │ 2018-09-10T22:00:00 │ 13.85       │ 83.0855  │
+│ 24  │ 2018-09-10T23:00:00 │ 12.75       │ 86.9574  │
+```
+
+We want to create some data features based on the time of day. One way to do this is with the `Periodic` transform, specifying a period of 1 day:
+
+```jldoctest example
+julia> df[:hour_of_day_sin] = FeatureTransforms.apply(df, Periodic(sin, Day(1)); cols=:time);
+
+julia> feature_df = df
```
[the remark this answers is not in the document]
glennmoy: ok, not sure what's going on. Has it anything to do with using `jldoctest` here but defining `df` in a `julia-repl` block? or are those compatible?

docs/src/examples.md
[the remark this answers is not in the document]
glennmoy: yeah on second though those should fine.

docs/src/transforms.md
```diff
@@ -0,0 +1,246 @@
...
+ 1.0
+ 4.0
+ 9.0
+```
+
+Alternatively, the data can be mutated using the `apply!` method.
+
+!!! note
+
+    Not all `Transform` subtypes support mutation.
```
glennmoy, line 60:
```suggestion
    Not all `Transform` subtypes support mutation such as those which change the type or dimension of the input.
```

docs/src/transforms.md
```diff
@@ -0,0 +1,246 @@
...
+ 9.0
+
+julia> x
+3-element Array{Float64,1}:
+ 1.0
+ 4.0
+ 9.0
+```
+
+A single `Transform` instance can be applied to different data types, with support for `AbstractArray`s and [`Table`s](https://github.com/JuliaData/Tables.jl).
```
glennmoy, line 76: this reads like it should be the first sentence in the next section?

docs/src/examples.md
```diff
@@ -0,0 +1,139 @@
...
+
+julia> output_cols = [:temperature, :humidity];
+```
+
+For many models it is helpful to normalize the training data.
+We can use `MeanStdScaling` for that purpose.
+Note that we are mutating the data frame in-place using `apply!`, and the order of columns specified does not matter.
+
+```jldoctest example
+julia> scaling = MeanStdScaling(train_df; cols=input_cols);
```
glennmoy, line 88: Isn't it kinda weird to normalise a periodic signal? since it won't be normally distributed and the original Periodic transform already "scaled" it.

If you think so too, maybe just scale the `output_cols`?

glennmoy (approved):

docs/src/examples.md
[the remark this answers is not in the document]
glennmoy: Just to note again this looks weird, if you can figure it out. But if not don't sweat it.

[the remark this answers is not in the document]
bencottier: Yeah it's not ideal but I'll leave it. The `julia-repl` block was changed, it's all `jldoctest` now.

docs/src/transforms.md
```diff
@@ -10,77 +11,199 @@ DocTestSetup = quote
...
+ 9.0
+```
 
-* `Transforms.apply(data, t; kwargs...)` is non-mutating, returning the transformed data without modifying the original data.
-* `t(data; kwargs...)` is equivalent to `Transforms.apply(data, t; kwargs...)`
-* `Transforms.apply!(data, t; kwargs...)` is mutating, returning the modified `data` with the same type.
+Alternatively, the data can be mutated using the `apply!` method.
 
-A single `Transform` can be applied to different data types and in different ways. The two main data types supported are `AbstractArray`s and [`Table`s](https://github.com/JuliaData/Tables.jl).
+!!! note
```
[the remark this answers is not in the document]
bencottier: I've seen it used this way elsewhere e.g. https://dataframes.juliadata.org/stable/man/getting_started/, and it renders: https://invenia.github.io/FeatureTransforms.jl/previews/PR37/transforms/#about-transforms

docs/src/transforms.md
bencottier: I wanted to lead into the next section with this, but for a user manual it's probably better to put it in the next section.

bencottier: Also it's talking about both arrays and tables, and the next section is just arrays.

docs/src/examples.md
bencottier: It doesn't matter much here because the periodic is roughly the same range of values anyway, but I'll just explain my view on normalising.

My logic is that all features should be normalised consistently, so that values lie within a similar small range around zero. In turn, model parameters require smaller updates. I think that mostly matters for neural nets with gradient descent. I'm not thinking as much about what the distribution of the feature is.

docs/src/transforms.md
bencottier: Leaving this for now.

Issue #38, opened 2021-03-02
Implement apply method that appends result to data and / or forces promotion


glennmoy:
Related to #12 

We have two methods for applying a transform to data
* `apply` takes the transform but preserves the original `data`
* `apply!` takes the transform and mutates the original `data` in-place

While `apply` is universally supported, `apply!` is only supported for transforms that can directly replace the input.

In one case, this means it needs the output to be the same type:

```julia
julia> p = Power(1.2);

julia> x = Int64[1, 2, 3];

julia> FeatureTransforms.apply(x, p)
3-element Array{Float64,1}:
 1.0
 2.2973967099940698
 3.7371928188465517

julia> FeatureTransforms.apply!(x, p)
ERROR: InexactError: Int64(2.2973967099940698)
```

In this example, we might just want to force the type promotion. 
But another simple case arises when the output is a different shape to the input.
Consider `LinearCombination`, which typically takes more than 1 input but produces just 1 output:

```julia
lc = LinearCombination([1, -1])
A = [1 2; 5 9]

julia> FeatureTransforms.apply(A, lc);  # works

julia> FeatureTransforms.apply!(A, lc)
ERROR: DimensionMismatch("tried to assign 2 elements to 4 destinations")
```

Note that this kind of transform is Many-to-One, so would expect similar problems for One-to-Many and Many-to-Many.

We therefore might want some `apply`-like methods that would:
1. Force mutation of the input (where possible) for One-to-One transforms by converting the underlying types.
2. Append the input with the result (where possible) for Many-to-One, One-to-Many, or Many-to-Many transforms.

Given the types of problems these are solving it might be desirable to have these achieved by separate methods.
But note that it's possible to solve both problems using (2) and this would be a consistent behaviour.

Here are some ideas for how we might approach the solution:
1. Special keyword args: `apply!(...; force=true)`, `apply!(...; append=true)`.
2. Special methods: `apply_force!`, `apply_and_append!`, also `apply!!` (cf https://github.com/JuliaFolds/BangBang.jl)
3. Define traits based on the transform [cardinality](https://en.wikipedia.org/wiki/Cardinality_(data_modeling)) with special rules in place for, e.g., `apply!!(x, ::OnetoOne; kwargs...`), `apply!!(x, ::ManytoOne; kwargs...)`, `apply!!(x, ::ManytoMany; kwargs...)`.

This also opens the question of how to name the columns for the appended data for a Table.
Should it be provided by the user? or automatically generated?

[...]

PR #40, opened 2021-03-02
RFC: Implement appending apply function

glennmoy:
POC for #38 

A few things to note
* I don't like the idea of this being a separate transform. I think it would just create extra steps to do exactly what we're doing here.
* I'm not against naming the method `apply_and_append` (or similar). I chose `apply!!` because of its brevity and I think it can be regarded as a reasonable extension to `apply` and `apply!`. Points against is that it might be easy to confuse?
* Case in point, the `!!` syntax would imply that the input is mutated directly. Right now it isn't but we might be able to do this if we want (I would be in favour given the above).
* This doesn't (directly) solve the issue of type promotion. I think, given how `apply!` is going to be used it might make sense for it to be able to do that? (I'm not aware of any implicit understand behind `!`-functions that preclude doing this?)

```julia

julia> p = Power(3);

julia> FeatureTransforms.apply!!(df, p; cols=[:a, :b], new_cols=[:a3, :b3])
5×4 DataFrame
 Row │ a      b      a3     b3    
     │ Int64  Int64  Int64  Int64 
─────┼────────────────────────────
   1 │     1      5      1    125
   2 │     2      4      8     64
   3 │     3      3     27     27
   4 │     4      2     64      8
   5 │     5      1    125      1

julia> M = reshape(1:9, 3, 3)
3×3 reshape(::UnitRange{Int64}, 3, 3) with eltype Int64:
 1  4  7
 2  5  8
 3  6  9

julia> FeatureTransforms.apply!!(M, p; dims=1, inds=1)
4×3 Array{Int64,2}:
 1   4    7
 2   5    8
 3   6    9
 1  64  343
```

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -112,3 +112,29 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
 
     return table
 end
+
+"""
+    apply!!(A::AbstractArray, ::Transform; dims=1, kwargs...)
+
+Applies the [`Transform`](@ref) to each element of `A` and appends the result to `A`, if
+possible, creating a new array.
+"""
+function apply!!(data::AbstractArray, t; dims=1, kwargs...)
+    other_dims = setdiff(1:length(size(data)), dims)
+    output = FeatureTransforms.apply(data, t; dims=1, kwargs...)
+    return cat(data, output; dims=other_dims)
+end
+
+"""
+    apply!!(table::T, ::Transform; col=nothing, new_names)::T where T
+
+Applies the [`Transform`](@ref) to each of the specified columns in the `table` and appends
+the result into a new table, if possible, with the given `new_names`.
+If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
+"""
+function apply!!(table::T, t; new_names, kwargs...) where T
+    Tables.istable(table) || throw(MethodError(apply, (table, t)))
+    result = NamedTuple{Tuple(new_names)}(apply(table, t; kwargs...))
+    table = merge(Tables.columntable(table), result)
+    return T <: NamedTuple ? table : T(table)
+end
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -119,23 +119,27 @@ end
 Applies the [`Transform`](@ref) to each element of `A` and appends the result to `A`, if
 possible, creating a new array.
 """
-function apply!!(data::AbstractArray, t; dims, kwargs...)::AbstractArray
-    output = apply(data, t; dims=dims, kwargs...)
+#TODO: syntax would imply this is mutating. Should we make it so a new array isn't constructed?
+function apply!!(data::AbstractArray, t; dims, inds=Colon(), kwargs...)::AbstractArray
+    output = apply(data, t; dims=dims, inds=inds, kwargs...)
     new_size = collect(size(data))
-    setindex!(new_size, 1, dims)
+    inds == Colon() || setindex!(new_size, 1, dims)  # no need to reshape if all inds are used
     return cat(data, reshape(output, new_size...); dims=dims)
 end
 
 """
-    apply!!(table, ::Transform; col=nothing, new_names)
+    apply!!(table, ::Transform; col=nothing, new_cols)
 
 Applies the [`Transform`](@ref) to each of the specified columns in the `table` and appends
-the result into a new table, if possible, with the given `new_names`.
+the result into a new table, if possible, with the given `new_cols`.
 If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
 """
-function apply!!(table, t; new_names, kwargs...)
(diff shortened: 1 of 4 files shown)


src/transformers.jl
```diff
@@ -145,3 +145,29 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
...
+
+Applies the [`Transform`](@ref) to each of the specified columns in the `table` and appends
+the result into a new table, if possible, with the given `new_names`.
+If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
+"""
+function apply!!(table::T, t; new_names, kwargs...) where T
+    Tables.istable(table) || throw(MethodError(apply, (table, t)))
+    result = NamedTuple{Tuple(new_names)}(apply(table, t; kwargs...))
+    table = merge(Tables.columntable(table), result)
+    return T <: NamedTuple ? table : T(table)
```
glennmoy, line 172: not happy about this part - would be nice if `NamedTuple` constructor were idempotent

src/transformers.jl
```diff
@@ -145,3 +145,29 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
 
     return table
 end
+
+"""
+    apply!!(A::AbstractArray, ::Transform; dims=1, kwargs...)
+
+Applies the [`Transform`](@ref) to each element of `A` and appends the result to `A`, if
+possible, creating a new array.
```
glennmoy, line 153: >creating a new array

I'd ideally have liked this to be mutating but it appending doesn't really work like that AFAIK

src/apply.jl
```diff
@@ -112,3 +112,34 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
...
     return table
 end
+
+"""
+    apply!!(A::AbstractArray, ::Transform; dims=1, kwargs...)
+
+Applies the [`Transform`](@ref) to each element of `A` and appends the result to `A`, if
+possible, creating a new array.
+"""
+#TODO: syntax would imply this is mutating. Should we make it so a new array isn't constructed?
```
nicoleepp, line 122: I am ok using `!!` but if we do we should try to follow the convention for what this already means in the wild, which I believe is mutating if possible, and if not it creates a new type, doing type promotions if needed

https://juliafolds.github.io/BangBang.jl/dev/

glennmoy: If we did use `apply!!` to do that, what syntax would you suggest for and "apply and append" method?

nicoleepp: I dont think it excludes appending?

glennmoy: I'm confused...the behaviour that BangBang follows doesn't do the "appending" step automatically AFAIK. It does type promotion instead (which is not exactly what we're looking for here)

nicoleepp: I'm partial to just adding a kwarg `append=true`

glennmoy: Just to note there are more [use-cases](https://github.com/mcabbott/AxisKeys.jl/blob/master/src/push.jl#L76-L83) that I found in the wild where it suggests that `!!` is "appending" the input.

glennmoy:
Just to note there are more [use-cases](https://github.com/mcabbott/AxisKeys.jl/blob/master/src/push.jl#L76-L83) that I found in the wild where it suggests that `!!` is "appending" the input.

src/apply.jl
glennmoy: It wouldn't be my preferred way of doing it but I can be convinced. Admittedly, it's because I'm biased against adding another keyword argument (c.f. `name`) and I quite like the elegance of `apply!!`.

src/apply.jl
```diff
@@ -112,3 +112,34 @@ function apply!(table::T, t::Transform; cols=nothing, kwargs...)::T where T
...
+function apply!!(data::AbstractArray, t; dims, inds=Colon(), kwargs...)::AbstractArray
+    output = apply(data, t; dims=dims, inds=inds, kwargs...)
+    new_size = collect(size(data))
```
bencottier, line 125:
```suggestion
    append_size = collect(size(data))
```
I was confused thinking this was the new size of the whole output

src/apply.jl
bencottier: To me, including the append in `apply!!` is _bending_ the interface, because it's this extra thing that's not explicit in the BangBang doc, and I can't find a function in BangBang that's analogous. (Any relevant examples out side BangBang.jl ?). But it's not clearly breaking it either.

I also think there's reason to provide a definitely-not-mutating apply-and-append, so e.g. we have `apply_append` and also `apply_append!!`.

glennmoy: Yeah I've cooled off the `apply!!` syntax. If you check slack you'll see that it comes with an implicit understanding of what it does so in this case we'd be deviating from that. I'm going with an `apply_append` method instead.

glennmoy:
Closing in favour of https://github.com/invenia/FeatureTransforms.jl/pull/69

PR #42, opened 2021-03-05
Refactor scaling to only compute one mean and std

### LINKED ISSUE #39: No way to apply scaling to certain slices of an array
MWE:

```julia
julia> M = reshape(collect(1:6), 3, 2)
3×2 Array{Int64,2}:
 1  4
 2  5
 3  6
```

We can apply `Power` to the second column:

```julia
julia> FeatureTransforms.apply(M, Power(2); dims=2, inds=[2])
3×1 Array{Int64,2}:
 16
 25
 36
```

because `dims=2` iterates each row and `inds=[2]` indexes the second element of each row.

But to normalize the second column with `MeanStdScaling`, we need to use `MeanStdScaling(M; dims=1)` (i.e. take the mean and std of each column). Then I see no way to only normalise the second column, because `dims` needs to be consistent.

```julia
julia> scaling = MeanStdScaling(M; dims=1)
MeanStdScaling((1 = 2.0, 2 = 5.0), (1 = 1.0, 2 = 1.0))

julia> FeatureTransforms.apply(M, scaling; dims=1, inds=[2])
1×2 Array{Float64,2}:
 0.0  0.0
```

Maybe we need a `slices` argument to only apply to certain slices?

glennmoy:
## Proposed Solution

When constructing `MeanStdScaling` it should compute a _single_ mean and std pair for all the data it is given and no longer compute it on a slice-by-slice basis.

This PR:
* Simplifies the code in `scaling.jl`
* Simplifies the constructor and formats it consistently no matter what data it was given
* Simplifies the parent `apply` functions (which no longer need to pass `name` downstream)
* De-couples the `MeanStdScaling` from the data, allowing it operate as an independent `Transform` like any other.
* Makes it's application more consistent with other transforms, i.e. "apply the transform to all the data it's given" rather than assuming it needs to be applied slice-by-slice (which no other transform does).

Cons:
* Now need to construct separate `MeanStdScaling` transforms for each column if you wish to use more than one. It's possible we can make a helper util for this in future.

## Problem Description

There are a few peculiarities with the way the `MeanStdScaling` transform is implemented that has given rise to some code smell.

1. We parse  a `name` kwarg in the parent `apply` method that only applies to this transform, which complicates the code in `apply` and gives a false impression that `name` is universal.
2. Keeping track of the names in `scaling` is also complicated and the way the constructor is handled differs whether it was computed using an array or table, and if it used all the data or not.
3. `MeanStdScaling` is not an independent transform; it is coupled directly to the data that created it (one cannot rename a column afterwards for example).
4. How it's used is also inconsistent with our other transforms (see example below)
5. Finally, using `mapslices` may not be the best idea after all (https://github.com/invenia/FeatureTransforms.jl/issues/18#issuecomment-786877479) and AxisArrays doesn't natively support it (it also doesn't look like anyone's in a rush to put it in https://github.com/JuliaArrays/AxisArrays.jl/pull/195) but refactoring that part was impossible without simplifying this first.

On point 3. Let's consider how a more simple transform, e.g. `Power`, operates
```julia
julia> A = [1 2; 5 9; 8 10];

julia> p = Power(3);
Power(3)
```

The default behaviour is that transform is applied to all of the data.
More specifically, the parameter `p.exponent` is applied uniformly to all of the data it is given.
We don't compute /apply a separate exponent to each slice (hint hint).
Although we can restrict it to certain slices of the data if we want.
```julia
julia> FeatureTransforms.apply(A, p)
3×2 Array{Int64,2}:
   1     8
 125   729
 512  1000

julia> FeatureTransforms.apply(A, p; dims=1, inds=1)
1×2 Array{Int64,2}:
 1  8
```

Compare this with `MeanStdScaling`, which is special because the constructor needs to first parse the data to create the transform and so the parameters will depend on what we give it.
But the same principle of "apply the same parameters to all the data" should still hold.

Looking at how we pass in the data, the first problem is that it can change how the constructor is represented, which makes it difficult to wrangle in the code.
Moreover, it's not obvious what these parameters mean.
```julia
julia> scaling_all = MeanStdScaling(A)
MeanStdScaling((all = 5.833333333333333,), (all = 3.7638632635454052,))

julia> scaling_selected = MeanStdScaling(A; dims=1)
MeanStdScaling((1 = 4.666666666666667, 2 = 7.0), (1 = 3.5118845842842465, 2 = 4.358898943540674))
```

Second, it makes the assumption that we want separate scaling parameters for each slice.
(This feature probably comes from my [earlier suggestion](https://github.com/invenia/FeatureTransforms.jl/issues/3#issuecomment-777655342) but that comment should have been clearer to prioritise simplicity.)

What this means is that when applying this to `A`, different elements are transformed by different variables, which violates the principle above.

```julia
julia> FeatureTransforms.apply(A, scaling_all)
3×2 Array{Float64,2}:
 -1.28414   -1.01846
 -0.221404   0.841334
  0.57565    1.10702

julia> FeatureTransforms.apply(A, scaling_selected; dims=1)
3×2 Array{Float64,2}:
 -1.04407    -1.14708
  0.0949158   0.458831
  0.949158    0.688247
```

Finally, the transform is overly-coupled to the data it was computed on. 
While it should have a memory of this data (as it is stateful) that memory lies in the scaling parameters themselves, after computing these it should operate as an independent transform like any other.
But right now, applying it to certain `dims`, `inds` and `cols` relies too heavily on the original data and it is therefore too brittle and restrictive.

Compare with this branch.
The scaling is computed across all data it is given, `dims` doesn't matter unless `inds` is provided.
And it returns a consistent constructor now matter what format it's in.
```julia
julia> A = [1 2; 5 9; 8 10];

julia> scaling = MeanStdScaling(A)
MeanStdScaling(5.833333333333333, 3.7638632635454052)

julia> scaling = MeanStdScaling(A; dims=1)
MeanStdScaling(5.833333333333333, 3.7638632635454052)

julia> scaling = MeanStdScaling(A; dims=1, inds=1)
MeanStdScaling(1.5, 0.7071067811865476)
```

Afterwards, the `scaling` is an independent transform like any other and can be applied freely.
This takes into account the data may have changed (via other transforms) so the user should be able to specify what to apply it to.
```julia
julia> FeatureTransforms.apply(A, scaling)
3×2 Array{Float64,2}:
 -0.707107   0.707107
  4.94975   10.6066
  9.19239   12.0208

julia> FeatureTransforms.apply(A, scaling; dims=1, inds=[2])
1×2 Array{Float64,2}:
 4.94975  10.6066
```

I'll note that I think this also closes #39 in that it transforming the second column is consistent. 
This is thanks to `scaling` now acting like an independent transform.
```julia
julia> M = reshape(collect(1:6), 3, 2);

#normalise the second column
julia> scaling = MeanStdScaling(M; dims=2, inds=2)
MeanStdScaling(5.0, 1.0)

julia> FeatureTransforms.apply(M, scaling; dims=2, inds=[2])
3×1 Array{Float64,2}:
 -1.0
  0.0
  1.0
```

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -83,9 +83,13 @@ We can use `MeanStdScaling` for that purpose.
 Note that we are mutating the data frame in-place using `apply!`, and the order of columns specified does not matter.
 
 ```jldoctest example
-julia> scaling = MeanStdScaling(train_df; cols=output_cols);
+julia> temp_scaling = MeanStdScaling(train_df; cols=[:temperature]);
 
-julia> FeatureTransforms.apply!(train_df, scaling; cols=output_cols)
+julia> hum_scaling = MeanStdScaling(train_df; cols=[:humidity]);
+
+julia> FeatureTransforms.apply!(train_df, temp_scaling; cols=[:temperature]);
+
+julia> FeatureTransforms.apply!(train_df, hum_scaling; cols=[:humidity])
 22×4 DataFrame
  Row │ time                 temperature  humidity     hour_of_day_sin
      │ DateTime             Float64      Float64      Float64
@@ -112,7 +116,9 @@ julia> FeatureTransforms.apply!(train_df, scaling; cols=output_cols)
 We can use the same `scaling` transform to normalize the test data:
 
 ```jldoctest example
-julia> FeatureTransforms.apply!(test_df, scaling; cols=output_cols)
+julia> FeatureTransforms.apply!(test_df, temp_scaling; cols=[:temperature]);
+
+julia> FeatureTransforms.apply!(test_df, hum_scaling; cols=[:humidity])
 2×4 DataFrame
  Row │ time                 temperature  humidity  hour_of_day_sin
      │ DateTime             Float64      Float64   Float64
@@ -127,7 +133,9 @@ We can scale this back to the original units of temperature and humidity by conv
 ```jldoctest example
 julia> predictions = DataFrame([-0.36 0.61; -0.45 0.68], output_cols);
 
-julia> FeatureTransforms.apply!(predictions, scaling; cols=output_cols, inverse=true)
+julia> FeatureTransforms.apply!(predictions, temp_scaling; cols=[:temperature], inverse=true);
+
+julia> FeatureTransforms.apply!(predictions, hum_scaling; cols=[:humidity], inverse=true)
 2×2 DataFrame
  Row │ temperature  humidity 
      │ Float64      Float64  
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -80,16 +80,16 @@ julia> output_cols = [:temperature, :humidity];
 
 For many models it is helpful to normalize the training data.
 We can use `MeanStdScaling` for that purpose.
-Note that we are mutating the data frame in-place using `apply!`, and the order of columns specified does not matter.
+Note that we are mutating the data frame in-place using `apply!` one column at a time.
 
 ```jldoctest example
-julia> temp_scaling = MeanStdScaling(train_df; cols=[:temperature]);
+julia> temp_scaling = MeanStdScaling(train_df; cols=:temperature);
 
-julia> hum_scaling = MeanStdScaling(train_df; cols=[:humidity]);
+julia> hum_scaling = MeanStdScaling(train_df; cols=:humidity);
 
-julia> FeatureTransforms.apply!(train_df, temp_scaling; cols=[:temperature]);
+julia> FeatureTransforms.apply!(train_df, temp_scaling; cols=:temperature);
 
-julia> FeatureTransforms.apply!(train_df, hum_scaling; cols=[:humidity])
+julia> FeatureTransforms.apply!(train_df, hum_scaling; cols=:humidity)
 22×4 DataFrame
  Row │ time                 temperature  humidity     hour_of_day_sin
      │ DateTime             Float64      Float64      Float64
@@ -116,9 +116,9 @@ julia> FeatureTransforms.apply!(train_df, hum_scaling; cols=[:humidity])
 We can use the same `scaling` transform to normalize the test data:
 
 ```jldoctest example
-julia> FeatureTransforms.apply!(test_df, temp_scaling; cols=[:temperature]);
+julia> FeatureTransforms.apply!(test_df, temp_scaling; cols=:temperature);
 
-julia> FeatureTransforms.apply!(test_df, hum_scaling; cols=[:humidity])
+julia> FeatureTransforms.apply!(test_df, hum_scaling; cols=:humidity)
 2×4 DataFrame
  Row │ time                 temperature  humidity  hour_of_day_sin
      │ DateTime             Float64      Float64   Float64
@@ -133,9 +133,9 @@ We can scale this back to the original units of temperature and humidity by conv
 ```jldoctest example
 julia> predictions = DataFrame([-0.36 0.61; -0.45 0.68], output_cols);
 
-julia> FeatureTransforms.apply!(predictions, temp_scaling; cols=[:temperature], inverse=true);
+julia> FeatureTransforms.apply!(predictions, temp_scaling; cols=:temperature, inverse=true);
 
-julia> FeatureTransforms.apply!(predictions, hum_scaling; cols=[:humidity], inverse=true)
+julia> FeatureTransforms.apply!(predictions, hum_scaling; cols=:humidity, inverse=true)
 2×2 DataFrame
  Row │ temperature  humidity 
      │ Float64      Float64  
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -78,8 +78,8 @@ A single `Transform` instance can be applied to different data types, with suppo
 !!! note
 
     Some `Transform` subtypes have restrictions on how they can be applied once constructed.
-    For instance, `MeanStdScaling` stores the mean and standard deviation of some data for specified dimensions or column names.
-    So `MeanStdScaling` should only be applied to the same data type and for the same dimensions or subset of column names specified in construction.
+    For instance, `MeanStdScaling` stores the mean and standard deviation of some data, potentially specified via some dimension and column names.
+    So `MeanStdScaling` should only be applied to the same data, and for the same dimension and subset of column names, as those used in construction.
 
 ## Applying to `AbstractArray`
 
@@ -116,9 +116,10 @@ julia> FeatureTransforms.apply(M, p; inds=[4, 5, 6])
 ### Applying along dimensions using `dims`
 
 Transforms can be applied to `AbstractArray` data with a `dims` keyword argument.
-This will apply the `Transform` to slices of the array along dimensions determined by `dims`.
-For example, given a `Matrix`, `dims=1` applies to each column, and `dims=2` applies
-to each row.
+This will apply the `Transform` to slices of the array along this dimension, which can be selected by the `inds` keyword.
+So when `dims` and `inds` are used together, the `inds` change from being the global indices of the array to the relative indices of each slice.
+
+For example, given a `Matrix`, `dims=1` slices the data column-wise and `inds=[2, 3]` selects the 2nd and 3rd rows.
 
 !!! note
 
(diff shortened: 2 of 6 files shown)


test/one_hot_encoding.jl
```diff
@@ -35,9 +35,6 @@
             @test FeatureTransforms.apply(x, ohe; inds=2:4) == [0 0 1; 0 1 0; 0 0 1]
             @test FeatureTransforms.apply(x, ohe; dims=:) == expected
 
-            @test_throws DimensionMismatch FeatureTransforms.apply(x, ohe; dims=1)
```
bencottier, line 38: Why are these removed?

test/scaling.jl
```diff
@@ -274,32 +279,33 @@
...
             end
 
             @testset "inds" begin
                 scaling = MeanStdScaling(M)
                 @test FeatureTransforms.apply(M, scaling; inds=[2, 3]) ≈ [-0.559017, -1.11803] atol=1e-5
                 @test FeatureTransforms.apply(M, scaling; dims=:, inds=[2, 3]) ≈ [-0.559017, -1.11803] atol=1e-5
 
-                scaling = MeanStdScaling(M; dims=1)
-                @test FeatureTransforms.apply(M, scaling; dims=1, inds=[2]) ≈ [0.0 0.707107 0.707107] atol=1e-5
+                scaling = MeanStdScaling(M; dims=1, inds=[2])
```
bencottier, line 295: Interesting that an unwrapped index now works:

```julia
julia> scaling = MeanStdScaling(M; dims=1, inds=[2])
MeanStdScaling(1.0, 1.0)

julia> scaling = MeanStdScaling(M; dims=1, inds=2)
MeanStdScaling(1.0, 1.0)

julia> FeatureTransforms.apply(M, scaling; dims=1, inds=[2])
1×3 Array{Float64,2}:
-1.0  0.0  1.0

julia> FeatureTransforms.apply(M, scaling; dims=1, inds=2)
3-element Array{Float64,1}:
 -1.0
  0.0
  1.0
```

The closest equivalent before (though the refactor and #39 make it not really comparable):

```julia
julia> scaling = MeanStdScaling(M; dims=2)
MeanStdScaling((1 = 0.0, 2 = 1.0), (1 = 0.5, 2 = 1.0))

julia> FeatureTransforms.apply(M, scaling; dims=2, inds=[2])
2×1 Array{Float64,2}:
 -1.0
  0.0

julia> FeatureTransforms.apply(M, scaling; dims=2, inds=2)
ERROR: MethodError: no method matching _apply(::Float64, ::MeanStdScaling; name=Symbol("1"))
Closest candidates are:
  _apply(::Any, ::Transform, ::Union{Tuple, AbstractArray}; kwargs...) at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:121
  _apply(::Any, ::Transform, ::Any; kwargs...) at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:117
  _apply(::Any, ::OneHotEncoding{R}; kwargs...) where R<:Real at /Users/bencottier/JuliaEnvs/Transform.jl/src/one_hot_encoding.jl:35
  ...
Stacktrace:
 [1] (::FeatureTransforms.var"#3#4"{Int64,Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},MeanStdScaling})(::Array{Float64,1}) at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:82
 [2] mapslices(::FeatureTransforms.var"#3#4"{Int64,Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},MeanStdScaling}, ::Array{Float64,2}; dims::Int64) at ./abstractarray.jl:2083
 [3] #apply#2 at /Users/bencottier/JuliaEnvs/Transform.jl/src/transformers.jl:80 [inlined]
 [4] top-level scope at REPL[5]:1
```

src/scaling.jl
```diff
@@ -12,93 +12,65 @@ Represents the no-op scaling which simply returns the `data` it is applied on.
...
+    # `AbstractArray` keyword arguments
+    * `dims=:`: the dimension along which to take the `inds` slices.
+    * `inds=:`: the slices of the array to use in computing the statistics. Default uses all slices.
+
+    # `Table` keyword arguments
+    * `cols=nothing`: the columns to use in computing the statistics. Default uses all columns.
+
+    !!! note
+        The keyword arguments must be specified the same way when you construct and [`apply`](@ref)
+        the transform, else the results will be inconsistent, or throw an error.
```
bencottier, line 49: So I guess the rationale of still having to specify `dims` and `inds` in `apply` is: we're providing the freedom to scale whatever data/dimension/slice the user wants.

Given that, I think this note should be rephrased to say that if you _want_ it to be consistent, you _should_ specify the same way.

src/scaling.jl
```diff
@@ -12,93 +12,65 @@ Represents the no-op scaling which simply returns the `data` it is applied on.
...
-        σ_pairs = [(Symbol(i), x) for (i, x) in enumerate(std(A; dims=dims))]
+    function MeanStdScaling(table; cols=nothing)
+        Tables.istable(table) || throw(MethodError(MeanStdScaling, table))
+        columntable = Tables.columns(table)
 
-        μ = (; μ_pairs...)
-        σ = (; σ_pairs...)
+        cols = _to_vec(cols)  # handle single column name
+        cnames = cols === nothing ? propertynames(columntable) : cols
+        data = reduce(vcat, [getproperty(columntable, c) for c in cnames])
```
bencottier, line 62: This is probably just because I'm set in the old way, but I'm concerned that the meaning of `cols` is confusing. I expect it to independently normalise the `cols` (which I think is a common use-case and a downside of this design). Instead it computes the stats over the whole block.

This behaviour is stated in the docstring but I think it's worth emphasising.

bencottier (commented):
I like this. It's cleaner, simpler, and more consistent with the `Transform` ethos.

However I have a couple of concerns:

- It's extra effort to create a scaling object for each slice or column. I think normalising multiple columns is a typical use case. Maybe we could add a convenience function for this?
- I think this only closes #39 with the change from `mapslices` to `selectdim`, not the refactor of scaling. The issue is about how `inds` plays with the `dims` convention.

By the way, the description is hard to follow when I don't already know the conclusion. I recommend the pyramid principle :) Summarise what this does and why, at the top.

test/one_hot_encoding.jl
glennmoy: had to do with removing `mapslices` - `selectdim` doesn't through a `DimensionMismatch`. It'll be in another PR

glennmoy:
> It's extra effort to create a scaling object for each slice or column. I think normalising multiple columns is a typical use case. 

Is it though? Arguably each column in a collection is its own feature, which would suggest creating separate transforms rather than doing them all once?

Feel free to post in slack with counter-examples from our code base. I'm happy to be corrected.

> I think this only closes #39 with the change from `mapslices` to `selectdim`, not the refactor of scaling. The issue is about how `inds` plays with the `dims` convention.

Fair enough. I compared at the last minute just out of curiosity.

> By the way, the description is hard to follow when I don't already know the conclusion. I recommend the pyramid principle :) Summarise what this does and why, at the top.

👍

src/scaling.jl
glennmoy: is there somewhere else you'd like to see it emphasised?

src/scaling.jl
glennmoy: reworded the docstring(s)

src/scaling.jl
bencottier: Just the docstring is fine

bencottier:
> Hmm...arguably each column in a collection is its own feature, which would suggest creating separate transforms rather than doing them all once? And I don't think we scale multiple columns in our own code base but I'm happy to be corrected if you want to post some examples in slack.

I have at least one example I'll post. 

That's a fair argument, but we allow applying to multiple columns at once for every other element-wise transform.

src/scaling.jl
```diff
@@ -12,93 +12,66 @@ Represents the no-op scaling which simply returns the `data` it is applied on.
 """
 struct IdentityScaling <: AbstractScaling end
 
-_apply(x, scaling::IdentityScaling; kwargs...) = x
-_apply!(x, scaling::IdentityScaling; kwargs...) = _apply(x, scaling; kwargs...)
-
+@inline _apply(x, ::IdentityScaling; kwargs...) = x
+@inline _apply!(x, ::IdentityScaling; kwargs...) = x
```
nicoleepp, line 16: Is this needed? Can we test this line if yes?

nicoleepp (approved):

src/scaling.jl
glennmoy: you're talking about the `_apply!` method - yeah we probably don't need it

docs/src/examples.md
```diff
@@ -83,9 +83,13 @@ We can use `MeanStdScaling` for that purpose.
 Note that we are mutating the data frame in-place using `apply!`, and the order of columns specified does not matter.
```
bencottier, line 83:
```suggestion
Note that we are mutating the data frame in-place using `apply!`.
```
Still true, but no longer relevant for this example because it's one column at a time?

docs/src/transforms.md
```diff
@@ -116,9 +116,10 @@ julia> FeatureTransforms.apply(M, p; inds=[4, 5, 6])
 ### Applying along dimensions using `dims`
 
 Transforms can be applied to `AbstractArray` data with a `dims` keyword argument.
-This will apply the `Transform` to slices of the array along dimensions determined by `dims`.
-For example, given a `Matrix`, `dims=1` applies to each column, and `dims=2` applies
-to each row.
+This will apply the `Transform` to slices of the array along this dimension, which can be selected by the `inds` keyword.
+So when `dims` and `inds` are used together, the `inds` change from being the global indices of the array to the relative indices of each slice.
+
+For example, given a `Matrix`, `dims=1` slices the data column-wise and `inds=[2, 3]` selects the 2nd and 3rd rows.
```
glennmoy, line 122: @bencottier does this make sense to you? and the changes to the example below?

bencottier: Yeah looks good

PR #46, opened 2021-03-09
Refactor OHE tests

glennmoy:
I found the tests difficult to edit earlier. There are references to `expected` scattered about but it's not clear for any one test what that variable is.


rofinn (approved):
Seems reasonable to me.

PR #50, opened 2021-03-11
Add is_transformable

### LINKED ISSUE #49: Define the transform interface
The idea around a "transform interface” is to make feature transformations composable, i.e. the output of one `Transform` should be valid input to another.

Feature engineering pipelines, which comprise a sequence of multiple `Transform`s and other steps, should obey the same principle and one should be able to add/remove subsequent `Transform`s without the pipeline breaking.

So the output of an end-to-end transform pipeline should itself be "transformable".

We have enforced this in `Transform`s by supporting certain input types, i.e. `AbstractArray`s and `Table`s, which produce other `AbstractArray`s and `Table`s.

We want to enforce this in the more generic `transform` API, which is expected to be overloaded for implementing pipelines but the exact method is an implementation detail for the user.
 Currently the  interface just comprises a [`transform`](https://github.com/invenia/FeatureTransforms.jl/blob/main/src/transformers.jl#L24) stub without any rules associated with it.
Our only requirement the return of `transform` is itself "transformable", i.e. an `AbstractArray` or `Table`.

I think is all that is required?
```julia
is_transformable(::AbstractArray)  = true
is_transformable(x) = Tables.istable(x)
```

nicoleepp:
Closes #49

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FeatureTransforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.2"
+version = "0.2.3"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/transformers.jl
+++ b/src/transformers.jl
@@ -9,6 +9,15 @@ abstract type Transform end
 # Make Transforms callable types
 (t::Transform)(x; kwargs...) = apply(x, t; kwargs...)
 
+"""
+    is_transformable(x)
+
+
+Determine if `x` is transformable, which means that you can apply any `transform` to `x `.
+Currently Tables and AbstractArrays are transformable.
+"""
+is_transformable(::AbstractArray)  = true
+is_transformable(x) = Tables.istable(x)
 
 """
     transform!(::T, data)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -15,6 +15,7 @@ using TimeZones
     include("power.jl")
     include("scaling.jl")
     include("temporal.jl")
+    include("transformers.jl")
 
     doctest(FeatureTransforms)
 end
--- a/src/FeatureTransforms.jl
+++ b/src/FeatureTransforms.jl
@@ -7,7 +7,7 @@ using Tables
 export HoD, LinearCombination, OneHotEncoding, Periodic, Power
 export IdentityScaling, MeanStdScaling, AbstractScaling
 export Transform
-export transform, transform!
+export is_transformable, transform, transform!
 
 include("utils.jl")
 include("transformers.jl")
--- a/test/transformers.jl
+++ b/test/transformers.jl
@@ -1,0 +1,16 @@
+@testset "is_transformable" begin
+
+    # Test that AbstractArrays and Tables are transformable
+    @test is_transformable([1, 2, 3, 4, 5])
+    @test is_transformable([1 2 3; 4 5 6])
+    @test is_transformable(AxisArray([1 2 3; 4 5 6], foo=["a", "b"], bar=["x", "y", "z"]))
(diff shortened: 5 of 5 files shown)


src/transformers.jl
```diff
@@ -9,6 +9,15 @@ abstract type Transform end
 # Make Transforms callable types
 (t::Transform)(x; kwargs...) = apply(x, t; kwargs...)
 
+"""
+    is_transformable(x)
+
+
```
glennmoy, line 15:
```suggestion
```

src/transformers.jl
```diff
@@ -9,6 +9,15 @@ abstract type Transform end
...
+Determine if `x` is transformable, which means that you can apply any `transform` to `x `.
+Currently Tables and AbstractArrays are transformable.
```
glennmoy, line 17:
```suggestion
Determine if `x` is both a valid input and output of any [`Transform`](@ref), which defines 
the [`transform`](@ref) interface.
Currently, all subtypes of `Table`s and `AbstractArray`s are transformable.
```

test/runtests.jl
```diff
@@ -15,6 +15,7 @@ using TimeZones
     include("power.jl")
     include("scaling.jl")
     include("temporal.jl")
+    include("transformers.jl")
```
glennmoy, line 18: this might be a good opportunity to break out the interface stuff (`transform`, `Transform`) from all the `apply` methods into two files. Or at least rename `transformers.jl` to `transform.jl` (the original naming was a clash when the package was called `Transform.jl`)

What do you think?

src/transformers.jl
glennmoy: The docstring for `transform` and `transform!` probably needs updating as well.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "FeatureTransforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.2.2"
+version = "0.2.3"
```
glennmoy, line 4: You don't think this is breaking?

glennmoy (commented):
Looks good, I think the docs might also need to be updated. There's a reference to in the [landing page](https://invenia.github.io/FeatureTransforms.jl/stable/#FeatureTransforms) to the `transform` method. I suggest we make a separate page about it which can basically be the same content as the issue https://github.com/invenia/FeatureTransforms.jl/issues/49

src/transformers.jl
bencottier: > which defines the transform interface

I don't know what `which` refers to - this function, `Transform`, or being a valid input and output of any `Transform`. Could clarify that.

test/runtests.jl
bencottier: I'd be happy with separate `transform.jl` and `apply.jl`.

Project.toml
bencottier: https://github.com/invenia/FeatureTransforms.jl/issues/49#issuecomment-797071952

morris25: We don't seem to use `is_transformable` to restrict anything in the code so it's definitely not breaking. If it were a trait or used in a test suite it might be.

src/transforms.jl
```diff
@@ -0,0 +1,45 @@
...
+"""
+abstract type Transform end
+
+# Make Transforms callable types
+(t::Transform)(x; kwargs...) = apply(x, t; kwargs...)
+
+"""
+    is_transformable(x)
+
+Determine if `x` is both a valid input and output of any [`Transform`](@ref), i. e. that it
```
bencottier, line 15:
```suggestion
Determine if `x` is both a valid input and output of any [`Transform`](@ref), i.e. that it
```

src/transforms.jl
```diff
@@ -0,0 +1,45 @@
+
```
bencottier, line 1: I prefer `transform.jl` for the filename because with `transforms.jl` I might expect subtypes of `Transform` to be defined here.

bencottier (approved):
Looks good to me

glennmoy (approved):

PR #61, opened 2021-03-22
Return consistent table type

### LINKED ISSUE #55: Non mutating apply should return a DataFrame if using `[[:x]]`

glennmoy:
Closes #55 

Alternative to #60 (closed)

This circumvents #55 altogether by returning the same type of output for a given input.
Right now this is guaranteed for  `AbstractArray` but _not_ for Tables. So to me it makes sense to actually support this rather than continuing to special-case tables and then special-casing what `cols` does (++complexity, ++inconsistency).

We were against this initially because we didn't want to come up with a naming scheme for the returned Table.
However, [`Tables.table`](https://tables.juliadata.org/stable/#Tables.table) allows for default `headers` which seem perfectly reasonable for us to extend as part of supporting Tables. We could even expose this kwarg to the user so they can pass in their own `header` names if they want.

Furthermore, we can reconstruct the original input type thanks to [`Tables.materializer`](https://tables.juliadata.org/stable/#Tables.materializer) (thanks @morris25 for bringing my attention to this!) which makes this much easier and is actually intended precisely for this purpose in applying transformations:

```
For a table input, return the "sink" function or "materializing" function that can take a Tables.jl-compatible table input 
and make an instance of the table type. This enables "transform" workflows that take table inputs, apply transformations, 
potentially converting the table to a different form, and end with producing a table of the same type as the original input. 
The default materializer is Tables.columntable, which converts any table input into a NamedTuple of Vectors.
```

Benchmarking: note that it's less performant than main probably because we are constructing Tables, but at least it's type-stable.
```julia
julia> using DataFrames, FeatureTransforms
[ Info: Precompiling FeatureTransforms [8fd68953-04b8-4117-ac19-158bf6de9782]

julia> df = DataFrame(rand(100, 100));

julia> p = Power(3);

# main
julia> @time FeatureTransforms.apply(df, p, cols=[:x1]);
  0.153022 seconds (399.12 k allocations: 21.648 MiB)
  0.000019 seconds (8 allocations: 1.203 KiB)

julia> @time FeatureTransforms.apply(df, p, cols=:x1);
  0.006716 seconds (7.00 k allocations: 427.577 KiB)
  0.000011 seconds (5 allocations: 992 bytes)

# this branch
julia> @time FeatureTransforms.apply(df, p, cols=[:x1]);
  0.398376 seconds (1.29 M allocations: 68.336 MiB, 18.14% gc time)
  0.000036 seconds (35 allocations: 5.328 KiB)

julia> @time FeatureTransforms.apply(df, p, cols=:x1);
  0.012736 seconds (8.49 k allocations: 513.706 KiB)
  0.000043 seconds (34 allocations: 5.234 KiB)

julia> @code_warntype FeatureTransforms.apply(df, p, cols=:a)
Variables
  #unused#::Core.Compiler.Const(FeatureTransforms.var"#apply##kw"(), false)
  @_2::NamedTuple{(:cols,),Tuple{Symbol}}
  @_3::Core.Compiler.Const(FeatureTransforms.apply, false)
  table::DataFrame
  t::Power
  kwargs...::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}
  cols::Symbol
  @_8::Symbol

Body::DataFrame
1 ─       Core.NewvarNode(:(kwargs...))
│   %2  = Base.haskey(@_2, :cols)::Core.Compiler.Const(true, false)
│         %2
│         (@_8 = Base.getindex(@_2, :cols))
└──       goto #3
2 ─       Core.Compiler.Const(:(@_8 = FeatureTransforms.nothing), false)
3 ┄       (cols = @_8)
│   %8  = (:cols,)::Core.Compiler.Const((:cols,), false)
│   %9  = Core.apply_type(Core.NamedTuple, %8)::Core.Compiler.Const(NamedTuple{(:cols,),T} where T<:Tuple, false)
│   %10 = Base.structdiff(@_2, %9)::Core.Compiler.Const(NamedTuple(), false)
│         (kwargs... = Base.pairs(%10))
│   %12 = FeatureTransforms.:(var"#apply#4")(cols, kwargs..., @_3, table, t)::DataFrame
└──       return %12

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,50 +20,70 @@ Load in the dependencies and construct some toy data.
 ```julia
 julia> using DataFrames, FeatureTransforms
 
-julia> df = DataFrame(:a=>[1, 2, 3, 4, 5], :b=>[5, 4, 3, 2, 1], :c=>[0, 1, 0, 1, 0])
+julia> df = DataFrame(:a=>[1, 2, 3, 4, 5], :b=>[5, 4, 3, 2, 1], :c=>[2, 1, 3, 1, 3])
 5×3 DataFrame
- Row │ a      b      c
-     │ Int64  Int64  Int64
+ Row │ a      b      c     
+     │ Int64  Int64  Int64 
 ─────┼─────────────────────
-   1 │     1      5      0
+   1 │     1      5      2
    2 │     2      4      1
-   3 │     3      3      0
+   3 │     3      3      3
    4 │     4      2      1
-   5 │     5      1      0
+   5 │     5      1      3
 ```
 
-We construct the transformations that we want to `apply` to the data, which can be non-mutating (`apply`) or mutating (`apply!`) if supported.
-Note that non-mutating transformations do not necessarily return the same type, even when applied to all the elements.
+Next, we construct the `Transform` that we want to `apply` to the data, which can either be non-mutating (`apply`) or mutating (`apply!`).
+All `Transforms` support the non-mutating `apply` method but any `Transform` that changes the type or dimension of the input does not support mutation.
+
+In either case, the return will be the same type as the input, so if you provide an `Array` you get back an `Array`, and if you provide a `Table` you get back a `Table`.
+Here we are working with a `DataFrame`, so the return will always be a `DataFrame`:
 ```julia
 julia> p = Power(3);
 
-julia> FeatureTransforms.apply(df, p; cols=[:a])
-1-element Array{Array{Int64,1},1}:
- [1, 8, 27, 64, 125]
+julia> FeatureTransforms.apply(df, p; cols=[:a], header=[:a3])
+5×1 DataFrame
+ Row │ a3    
+     │ Int64 
+─────┼───────
+   1 │     1
+   2 │     8
+   3 │    27
+   4 │    64
+   5 │   125
 
 julia> FeatureTransforms.apply!(df, p; cols=[:a])
 5×3 DataFrame
  Row │ a      b      c
      │ Int64  Int64  Int64
 ─────┼─────────────────────
-   1 │     1      5      0
+   1 │     1      5      2
    2 │     8      4      1
-   3 │    27      3      0
+   3 │    27      3      3
    4 │    64      2      1
-   5 │   125      1      0
+   5 │   125      1      3
 ```
 
-Also note that some transformations, such as those applying a reduction operation, do not support mutation.
-But users may append the output to their data if they so wish.
+
+`Transform`s that don't support mutation must be called using `apply` and appended.
+To help with this, you can call the `Transform` type directly:
 ```julia
+julia> ohe = OneHotEncoding(1:3);
+
 julia> lc = LinearCombination([1, -10]);
 
-julia> FeatureTransforms.apply(df, lc; cols=[:b, :c])
-5-element Array{Int64,1}:
-  5
- -6
-  3
- -8
-  1
+julia> ohe_df = ohe(df; cols=[:c], header=[:cat1, :cat2, :cat3])
+
+julia> lc_df = lc(df; cols=[:a, :b], header=[:ab]);
+
+julia> df = hcat(df, lc_df, ohe_df)
+5×7 DataFrame
+ Row │ a      b      c      ab     cat1   cat2   cat3  
+     │ Int64  Int64  Int64  Int64  Bool   Bool   Bool  
+─────┼─────────────────────────────────────────────────
+   1 │     1      5      2    -49  false   true  false
+   2 │     8      4      1    -32   true  false  false
+   3 │    27      3      3     -3  false  false   true
+   4 │    64      2      1     44   true  false  false
+   5 │   125      1      3    115  false  false   true
 
 ```
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -61,7 +61,7 @@ function apply!(A::AbstractArray, t::Transform; kwargs...)
 end
 
 """
-    apply(table, ::Transform; cols=nothing, kwargs...) -> Vector
+    apply(table, ::Transform; cols=nothing, kwargs...) -> Table
 
 Applies the [`Transform`](@ref) to each of the specified columns in the `table`.
 If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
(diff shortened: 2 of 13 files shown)


morris25:
I think this solution is preferable to #60 since we are only supporting DataFrames via the Tables interface.

bencottier:
> Furthermore, we can reconstruct the original input type thanks to Tables.materializer

This is new info for me and changes my preference. In the past I've wished we had this, and the array return type stuff has just been the best we can do.

This also changes the question on the single-column syntax. Before, with #26 , it was just in the context of what kind of array to return. If everything's a Table, I don't mind there being no distinction in this syntax.

Overall I like this and it feels more like how it's meant to be than #60

glennmoy:
> one caveat is that I'm still not sure what users would want here. Can we glean this from the test session?

There was nothing explicitly said about it. But I think the option to specify the column header is enough, as long as we fall back to default `Tables.table` then at least we are not asserting out opinion on it.

src/apply.jl
```diff
@@ -77,16 +77,23 @@ function apply(table, t::Transform; cols=nothing, kwargs...)
...
-    return _apply(columntable, t, cnames; kwargs...)
+
+    result = _apply(columntable, t, cnames; kwargs...)
+    header = get(kwargs, :header, _default_header(result))
+    return Tables.materializer(table)(Tables.table(result, header=header))
 end
 
+# TODO: any way to not have to copy this from Tables.jl?
+# https://github.com/JuliaData/Tables.jl/blob/34be1bc063e0e1b01c8920fdeee9c70387f023bb/src/matrix.jl#L54
+_default_header(x) = [Symbol(:Column, i) for i in 1:size(Tables.matrix(Tables.table(x)), 2)]
```
glennmoy, line 88: this annoyed me and I'd rather not have it around, any one know how I can get rid of it?

src/linear_combination.jl
```diff
@@ -61,10 +61,15 @@ function apply(table, LC::LinearCombination; cols=nothing)
 
     # Keep the generic form when not specifying column names
     # because that is much more performant than selecting each col by name
-    cols === nothing && return [_sum_row(row, LC.coefficients) for row in Tables.rows(table)]
+    result = if cols === nothing
+        _to_mat([_sum_row(row, LC.coefficients) for row in Tables.rows(table)])
```
glennmoy, line 65: might just use `hcat` directly (c.f. `apply.jl`)

glennmoy:
Note I haven't updated the docs yet - nor have I tested passing through a `header` kwarg except for the doctest.

test/linear_combination.jl
```diff
@@ -168,14 +168,9 @@
         @testset "all cols" begin
             nt = (a = [1, 2, 3], b = [4, 5, 6])
             lc = LinearCombination([1, -1])
-            @test FeatureTransforms.apply(nt, lc) == [-3, -3, -3]
-            @test lc(nt) == [-3, -3, -3]
-        end
-
-        @testset "dims not supported" begin
```
bencottier, line 175: Curious why this was removed

test/scaling.jl
```diff
@@ -464,112 +466,122 @@
...
-                @test scaling(nt; cols=[c]) ≈ [collect(nt_expected_[c])] atol=1e-14
+                expected = (Column1 = [0.0, -1.0, 1.0], )
+                @test FeatureTransforms.apply(nt, scaling; cols=:a) == expected
+                @test FeatureTransforms.apply(nt, scaling; cols=[:a]) == expected
+                @test scaling(nt; cols=:a) == expected
 
                 _nt = deepcopy(nt)
-                FeatureTransforms.apply!(_nt, scaling; cols=[c])
+                FeatureTransforms.apply!(_nt, scaling; cols=:a)
                 @test _nt isa NamedTuple{(:a, :b)}  # before applying `collect`
```
bencottier, line 501:
```suggestion
```
I think it's unnecessary due to change below

bencottier (commented):
I'm satisfied with the tests so far.

test/linear_combination.jl
glennmoy: `apply(data, ::LinearCombination, ...)` previously didn't have a `dims` kwarg so this returned `MethodError`.

But now since it has to be able to accept arbitrary `kwargs` (to accept `header`) this no longer occurs.

I couldn't set default `header` in the usually `apply` because it isn't know until the transform is applied how many cols the result will have. But in this case we do know (it'll be 1) so I could set `header` to be explicit.

test/scaling.jl
glennmoy: yep, a bunch of tests are now unnecessary

src/apply.jl
glennmoy: genius

src/linear_combination.jl
glennmoy: In the end I just got rid of `_to_mat`. The code is getting quite messy and confusing so I'm cleaning it up in a follow up PR #63

README.md
```diff
@@ -20,50 +20,71 @@ Load in the dependencies and construct some toy data.
...
+   3 │     3      3      3
    4 │     4      2      1
-   5 │     5      1      0
+   5 │     5      1      3
 ```
 
-We construct the transformations that we want to `apply` to the data, which can be non-mutating (`apply`) or mutating (`apply!`) if supported.
-Note that non-mutating transformations do not necessarily return the same type, even when applied to all the elements.
+Next, we construct the `Transform` that we want to `apply` to the data, which can either be non-mutating (`apply`) or mutating (`apply!`).
+All `Transforms` support the non-mutating `apply` method any `Transform` that changes the type or dimension of the input does not support mutation.
```
nicoleepp, line 36:
```suggestion
All `Transforms` support the non-mutating `apply` method but any `Transform` that changes the type or dimension of the input does not support mutation.
```

README.md
```diff
@@ -20,50 +20,71 @@ Load in the dependencies and construct some toy data.
...
+
+In either case, the return will be the same type as the input.
+So if you provide an `Array` you get back an `Array`, and if you provide a `Table` you will get back a `Table`.
```
nicoleepp, line 39: should probably be combined with the previous sentence.

In either case, the return will be the same type as the input, so if you provide ...

nicoleepp (approved):

PR #45, opened 2021-03-09
Replace mapslices with selectdim

glennmoy:
This PR replaces the call to `mapslices` with a call to `selectdim` instead.

We change this now because we no longer need to take slices in the parent `apply` method(s) (https://github.com/invenia/FeatureTransforms.jl/pull/42). 
So we can replace `mapslices` with `selectdim` instead, which is simpler and _much_ more performant (see below), yet keeps the desired `dims` convention.

## background
This is a follow up to https://github.com/invenia/FeatureTransforms.jl/pull/42, which removed the need for looping over the slices in `apply`, and https://github.com/invenia/FeatureTransforms.jl/pull/25, which was aimed at solving our `dims` convention https://github.com/invenia/FeatureTransforms.jl/issues/18.

We originally liked `mapslices` because we wanted to adopt its `dims` convention, which was more intuitive to our users.
However, it has some considerable downsides that make it a poor choice in the long-term, namely:
* Greater attention and long-term development going into `eachslice`
  *  https://github.com/JuliaLang/julia/pull/32310 introducing `SlicedArrays` and abstracting the `dims` argument.
* Much poorer performance compared to `eachslice` 
  * https://github.com/JuliaLang/julia/issues/26868
  * https://github.com/JuliaLang/julia/issues/26232
* Cannot return a `view` https://github.com/JuliaLang/julia/issues/29146
* No support in AxisArrays (but given the above there won't be a hurry to https://github.com/JuliaArrays/AxisArrays.jl/pull/195)

Its one technical advantage was allowing multiple `dims`, but this isn't a necessary feature right now.

Comparing to [previous benchmarks](https://github.com/invenia/FeatureTransforms.jl/pull/25#issuecomment-784184604) this significantly reduces allocations (x100) for element-wise transforms (although `MeanStdScaling` is a bit of an unfair comparison since it has been simplified in the meantime). It might be worth looking to make the rest more efficient also.

```julia
# Periodic - this branch
# apply
0.333721 seconds (1.46 M allocations: 70.581 MiB, 7.08% gc time)
0.000225 seconds (11 allocations: 234.734 KiB)
# apply!
0.014992 seconds (31.75 k allocations: 1.924 MiB)
0.000282 seconds (10 allocations: 234.703 KiB)

# Periodic - main
apply
0.780758 seconds (2.84 M allocations: 146.639 MiB, 5.11% gc time)
0.000626 seconds (1.64 k allocations: 380.922 KiB)
apply!
[...]

PR #62, opened 2021-03-22
Test on dimnames for KeyedArrays

### LINKED ISSUE #20: Support named dims for `dims` keyword  argument
To use a consistent convention for `dims` (#18), we currently have to "invert" the value passed into the `dims` keyword argument to the array-based `apply!`, so that `eachslice` gives the expected result. For example `dims=2` becomes `dims=1` and `dims=3` becomes `dims=[1, 2]`. Basically we can use `setdiff(1:ndims(array), dims)` to accomplish that.

The solution isn't so straightforward for named dims, because the method to get the list of dim names is not universal. `KeyedArray` supports `dimnames(A)` but `AxisArray` does not, instead using `axisnames`. One workaround is `dimnames(wrapdims(::AxisArray))` but that requires adding `AxisKeys` as a dependency. 

Some have suggested that support for `AxisArrays` should be dropped entirely, or included in an auxiliary package - then we could rely on `dimnames`.

glennmoy:
Closes #20 

I'll note that #20 describes the situation before we deprecated both `eachslice` (https://github.com/invenia/FeatureTransforms.jl/pull/15) and `mapslices` (https://github.com/invenia/FeatureTransforms.jl/pull/45) in the parent `apply` methods. So the problem it is describing is outdated since we no longer need to take the complement set of dimensions for any transformations.

Still it is important that we support `dimnames` since user feedback [suggested](https://github.com/invenia/FeatureTransforms.jl/issues/20#issuecomment-801334106) "Not having named dims limits the usefulness of using a KeyedArray as opposed to just an AbstractArray". Fortunately, we already support this, this PR is just making that explicit via the tests.

However
* You can't pass in the keys as `inds` (for now, we might support this in future)
* This doesn't work with AxisArrays, at least not without `AxisArrays` first supporting `selectdim`. Considering we failed to get [`mapslices`](https://github.com/JuliaArrays/AxisArrays.jl/pull/195) into AxisArrays I'm not in a rush to support it as we'll be dropping it internally altogether soon enough.

### DIFF
--- a/test/linear_combination.jl
+++ b/test/linear_combination.jl
@@ -143,11 +143,13 @@
 
             @testset "dims = 1" begin
                 @test FeatureTransforms.apply(A, lc; dims=1) == [-3, -3]
+                @test FeatureTransforms.apply(A, lc; dims=:foo) == [-3, -3]
                 @test lc(A; dims=1) == [-3, -3]
             end
 
             @testset "dims = 2" begin
                 @test FeatureTransforms.apply(A, lc; dims=2) == [-1, -1]
+                @test FeatureTransforms.apply(A, lc; dims=:bar) == [-1, -1]
                 @test lc(A; dims=2) == [-1, -1]
             end
         end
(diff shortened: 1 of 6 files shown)


bencottier (approved):
Looks good to me

PR #67, opened 2021-03-26
Update doctests for Julia 1.6

glennmoy:
Note that I've also updated the branch settings for `main` so that `CI / Julia 1 - ubuntu-latest - x64 (pull_request)` job must now pass. This is consistent with our other open source packages.

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -31,13 +31,13 @@ julia> p = Power(2);
 julia> x = [1.0, 2.0, 3.0];
 
 julia> FeatureTransforms.apply(x, p)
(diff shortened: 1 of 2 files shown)


glennmoy:
I don't need to update the Project.toml right?

https://github.com/invenia/FeatureTransforms.jl/blob/main/Project.toml#L17

bencottier (approved):

bencottier:
> I don't need to update the Project.toml right?
> 
> https://github.com/invenia/FeatureTransforms.jl/blob/main/Project.toml#L17

Don't think so, since doctest is just ignored rather than failing for 1.5

PR #69, opened 2021-03-30
Define apply_append method

glennmoy:
Closes https://github.com/invenia/FeatureTransforms.jl/issues/38

Supersedes https://github.com/invenia/FeatureTransforms.jl/pull/40

```julia

julia> p = Power(3);

julia> FeatureTransforms.apply_append(df, p; cols=[:a, :b], header=[:a3, :b3])
5×4 DataFrame
 Row │ a      b      a3     b3    
     │ Int64  Int64  Int64  Int64 
─────┼────────────────────────────
   1 │     1      5      1    125
   2 │     2      4      8     64
   3 │     3      3     27     27
   4 │     4      2     64      8
   5 │     5      1    125      1

julia> M = reshape(1:9, 3, 3)
3×3 reshape(::UnitRange{Int64}, 3, 3) with eltype Int64:
 1  4  7
 2  5  8
 3  6  9

julia> FeatureTransforms.apply_append(M, p; dims=1, inds=1, append_dim=1)
4×3 Array{Int64,2}:
 1   4    7
 2   5    8
 3   6    9
 1  64  343
```

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -101,3 +101,27 @@ function apply!(table::T, t::Transform; cols=_get_cols(table), kwargs...)::T whe
 
     return table
 end
+
+"""
+    apply_append(A::AbstractArray, ::Transform; append_dim, kwargs...)
+
+Applies the [`Transform`](@ref) to `A` and returns the result in a new array where the output
+is appended to `A` along the `append_dim` dimension. The remaining `kwargs` correspond to
+the usual [`Transform`](@ref) being invoked.
+"""
+function apply_append(A::AbstractArray, t; append_dim, kwargs...)::AbstractArray
+    return cat(A, apply(A, t; kwargs...); dims=append_dim)
+end
+
+"""
+    apply_append(table, ::Transform; [header], kwargs...)
+
+Applies the [`Transform`](@ref) to the `table` and appends the result in a new table with an
+optional `header`. If none is provided the default in `Tables.table` is used. The remaining
+`kwargs` correspond to the [`Transform`](@ref) being invoked.
+"""
+function apply_append(table, t; kwargs...)
+    T = Tables.materializer(table)
+    result = Tables.columntable(apply(table, t; kwargs...))
+    return T(merge(Tables.columntable(table), result))
+end
--- a/src/linear_combination.jl
+++ b/src/linear_combination.jl
@@ -28,7 +28,6 @@ function apply(
     return _sum_terms(eachslice(selectdim(A, dims, inds); dims=dims), LC.coefficients)
 end
 
-
 """
     apply(table, LC::LinearCombination; [cols], [header]) -> Table
 
@@ -48,6 +47,15 @@ function apply(table, LC::LinearCombination; cols=_get_cols(table), header=nothi
 
     result = hcat(_sum_terms([getproperty(coltable, col) for col in cols], LC.coefficients))
     return Tables.materializer(table)(_to_table(result, header))
+end
+
+function apply_append(
+    A::AbstractArray{<:Real, N}, LC::LinearCombination; append_dim, kwargs...
+)::AbstractArray{<:Real, N} where N
+    # A was reduced along the append_dim so we must reshape the result setting that dim to 1
+    new_size = collect(size(A))
+    setindex!(new_size, 1, dim(A, append_dim))
+    return cat(A, reshape(apply(A, LC; kwargs...), new_size...); dims=append_dim)
 end
 
 function _sum_terms(terms, coeffs)
--- a/test/linear_combination.jl
+++ b/test/linear_combination.jl
@@ -37,6 +37,12 @@
             @test FeatureTransforms.apply(x, lc) == fill(-.1)
             @test lc(x) == fill(-.1)
         end
+
+        @testset "apply_append" begin
+            x = [1, 2]
+            lc = LinearCombination([1, -1])
+            @test FeatureTransforms.apply_append(x, lc; append_dim=1) == [1, 2, -1]
+        end
     end
 
     @testset "Matrix" begin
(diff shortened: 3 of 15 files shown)


src/linear_combination.jl
```diff
@@ -50,6 +49,15 @@ function apply(table, LC::LinearCombination; cols=_get_cols(table), header=nothi
     return Tables.materializer(table)(_to_table(result, header))
 end
 
+function apply_append(
+    A::AbstractArray{<:Real, N}, LC::LinearCombination; append_dim, kwargs...
+)::AbstractArray{<:Real, N} where N
+    # A was reduced along the append_dim so we must reshape the result setting that dim to 1
+    new_size = collect(size(A))
+    setindex!(new_size, 1, dim(A, append_dim))
```
glennmoy, line 57: this was unfortunately necessary to introduce NamedDims as a dependency.

Without `NamedDims.dim` there was no way to map the dimname to the dim - and so no way to know how to reshape the output of LinearCombination so it could be `cat` to the input.

src/apply.jl
```diff
@@ -101,3 +101,27 @@ function apply!(table::T, t::Transform; cols=_get_cols(table), kwargs...)::T whe
...
+
+"""
+    apply_append(A::AbstractArray, ::Transform; append_dim, kwargs...)
+
+Applies the [`Transform`](@ref) to `A` and returns the result in a new array where the output
+is appended to `A` along the `append_dim` dimension. The remaining `kwargs` correspond to
+the usual [`Transform`](@ref) being invoked.
+"""
+function apply_append(A::AbstractArray, t; append_dim, kwargs...)::AbstractArray
+    return cat(A, apply(A, t; kwargs...); dims=append_dim)
```
bencottier, line 113: I guess this copies the axiskeys of the chunk being applied to, so there isn't the equivalent of `header`?

I can understand not needing this, given axis keys are more like metadata for each data point. For example even if you take `Power(3)` of the data, it still corresponds to `time=t`.

What's your thinking on it?

src/linear_combination.jl
bencottier: This comes down to `setindex!` though, right? Because it has to use an integer index. Just wondering if there is an alternative outside of `setindex!`.

src/apply.jl
```diff
@@ -101,3 +101,27 @@ function apply!(table::T, t::Transform; cols=_get_cols(table), kwargs...)::T whe
...
 end
+
+"""
+    apply_append(A::AbstractArray, ::Transform; append_dim, kwargs...)
+
+Applies the [`Transform`](@ref) to `A` and returns the result in a new array where the output
+is appended to `A` along the `append_dim` dimension. The remaining `kwargs` correspond to
+the usual [`Transform`](@ref) being invoked.
+"""
+function apply_append(A::AbstractArray, t; append_dim, kwargs...)::AbstractArray
```
bencottier, line 112: Looking at some tests repeating `dims=x, append_dim=x` made me think, would it make sense to set the default value for `append_dim` as `dims` inside the function? Maybe that only works when `dims` is single, as is currently the case.

bencottier (approved):

src/linear_combination.jl
glennmoy: >This comes down to setindex! though, right? Because it has to use an integer index.

yeah exactly. I'm not sure what alternative exists though. I don't think there's anything in base that assumes a dim can be a symbol.

test/linear_combination.jl
```diff
@@ -81,6 +87,17 @@
             @test FeatureTransforms.apply(M, lc; inds=[2, 3]) == [3, -2]
             @test lc(M; inds=[2, 3]) == [3, -2]
         end
+
+        @testset "apply_append" begin
+            M = [1 1 1; 2 2 2; 3 3 3]
+            lc = LinearCombination([1, 1, 1])
+
+            expected1 = [1 1 1; 2 2 2; 3 3 3; 6 6 6]
+            @test FeatureTransforms.apply_append(M, lc; dims=1, append_dim=1) == expected1
```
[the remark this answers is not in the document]
glennmoy: It's technically possible but for linear combination it only makes sense to append to the dim you reduced over. It's not guaranteed to work any other way so I don't think testing for it is needed.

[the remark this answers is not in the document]
glennmoy: "not guaranteed to work" not because the _code_ won't allow it, but because the structure of your data won't.

src/apply.jl
glennmoy: >I guess this copies the axiskeys of the chunk being applied to, so there isn't the equivalent of header?

Correct, I went looking in the AxisKeys code for something like it but [`new_keys` is set internally](https://github.com/mcabbott/AxisKeys.jl/blob/master/src/functions.jl#L172) and can't be over-written externally.

>What's your thinking on it?

For transforms that affect the whole (2D) Array, it makes sense to append on the 3rd dimension and create a new dim that way. Otherwise you end up with redundant keys in one dimension.

For linear combination, or transforms on single columns it seems unavoidable for now that you have to rekey on the other side. We could do that processing here? But it would require loading AxisKeys explicitly to allow new keys be provided. For now I'm happy to leave it and see how much of an issue it becomes.

src/apply.jl
glennmoy: It definitely does for linear combination... I'll have a look and see if it makes sense in general and make the change if so.

src/apply.jl
glennmoy: I think I'll leave this for another PR - I have an idea or two of how to make this work but it'll require another review

src/apply.jl
glennmoy: I think I'll leave this for another PR - I have an idea or two of how to make this work but it'll require another review

glennmoy: https://github.com/invenia/FeatureTransforms.jl/issues/70

Issue #75, opened 2021-04-07
Use traits to generalise apply methods


glennmoy:
## Summary 

We should define a `Cardinality` trait for `Transform`s and appropriate intermediate `_apply` method for each. We should also create dummy `Transform`s to verify that types supporting the `transform` interface are compatible with any kind of `Transform`.

## Challenges

There are a few challenges in developing this package in it's current state:
* `LinearCombination` is a special case in that it has its own `apply` method, rather than just a more simple `_apply` because of how it needs to modify the input data.
* This makes it more difficult to add another transform like `LinearCombination` in the future: it will also likely need its own `apply` method.
* We have been explicitly testing each `Transform` on all supported data types and thus over-testing the package, but we have had to do this to make sure we are catching all edge-cases.
* If we were to support a new data type, we would have to write new tests against all `Transforms` for the same reason, which is extremely time-consuming and error-prone. https://github.com/invenia/AxisSets.jl/pull/44
* All of this is especially needed for `LinearCombination` because of its unique behaviour.

We need a better way of structuring the package so that
* New transforms are easy to write and test: 
    * No more special `apply` methods.
    * Only the `_apply` method  and the behaviour of the transform should be tested, which should be independent of the data type it's being used on.
* New data types are easy to support and test: 
    * Only the generic, high-level `apply`/`apply!`/`apply_append` methods are required.
    * We don't want to test each transform individually: we should test some canonical transforms that represent any transform that can be passed.

## Idea: A Cardinality Trait

The reason `LinearCombination` needs its own `apply` method is that it is a reduction operation, i.e. many-to-one, whereas (most of) the rest of the transforms are one-to-one.
The only exception is `OneHotEncoding`, which is one-to-many and comes with other challenges (like how to treat matrix inputs).

If we inspect the code, the difference is most easily recognised in the `tables` methods:
```julia
# Generic Transform: apply transform to each component in turn and collect the result
reduce(hcat, [_apply(getproperty(coltable, col), t; kwargs...) for col in cols])

# LinearCombination: collect all components and apply the transform
hcat(_sum_terms([getproperty(coltable, col) for col in cols], LC.coefficients))
```

It may be useful to define some interstitial level of abstraction between the `apply` method (for types) and `_apply` method (for `Transform`s) that treats the data according to the _cardinality_ of the `Transform`.

The rough structure of that design would look like:
```julia
abstract type Cardinality end

struct OneToOne <: Cardinality end
[...]

PR #77, opened 2021-04-09
POC: Add traits to support generalising the apply methods

glennmoy:
POC for #75 

Some brief notes to guide review:

This started off implementing the idea in the description of #75: adding traits to describe the [cardinality](https://en.wikipedia.org/wiki/Cardinality_(data_modeling)) of our transforms and defining an intermediate `_apply` method that dispatched on the traits.

But after some refactoring at the end, I noticed that the key difference boiled down to how the data was _prepared_ before going into a `ManyToOne` transform (`LinearCombination`).

So I refactored a bit more to create two simple formatting functions:
* `_preformat` structures the input to `_apply` according to the cardinality of the transform. Effectively, this just calls `eachslice` before calling `_apply` for `LinearCombination`.
* `_preformat` structures the output of `_apply` according to the cardinality of the transform. This is really only needed for `append_apply` because `LinearCombination` returns a `Vector`, if we are appending it as a row `cat` throws a DimensionMismatch. So we have to pivot the data before doing so.

I added some `TestUtils` with `Fakes` for the above transforms and an example testset for how we might test a new data type in a more manageable way in the future.
The idea being that: once a datatype supports each type of transform (and returns correct result for the `kwargs`) then it should support _any_  of the transforms defined in the package.

Then, each transform just needs to be tested against `Array`s and nothing else, which should give us full test coverage but in a much simpler way.

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -41,13 +41,13 @@ specified.
 function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
     if dims === Colon()
         if inds === Colon()
-            return _apply(A, t; kwargs...)
+            return _apply(cardinality(t), A, t; dims=:, kwargs...)
         else
-            return @views _apply(A[:][inds], t; kwargs...)
+            return _apply(cardinality(t), A[:][inds], t; dims=:, kwargs...)
         end
     end
 
-    return _apply(selectdim(A, dims, inds), t; kwargs...)
+    return _apply(cardinality(t), selectdim(A, dims, inds), t; dims=dims, kwargs...)
 end
 
 """
@@ -59,6 +59,9 @@ function apply!(A::AbstractArray, t::Transform; kwargs...)
     A[:] = apply(A, t; kwargs...)
     return A
 end
+
+_apply(::Union{OneToOne, OneToMany}, A, t; kwargs...) = _apply(A, t; kwargs...)
+_apply(::Union{ManyToOne, ManyToMany}, A, t; dims, kwargs...) = _apply(eachslice(A; dims=dims), t; kwargs...)
 
 """
     apply(table, ::Transform; [cols], [header], kwargs...) -> Table
@@ -77,7 +80,8 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     coltable = Tables.columntable(table)
     cols = _to_vec(cols)
 
-    result = reduce(hcat, [_apply(getproperty(coltable, col), t; kwargs...) for col in cols])
+    components = reduce(hcat, getproperty(coltable, col) for col in cols)
+    result = hcat(_apply(cardinality(t), hcat(components), t; dims=2, kwargs...))
     return Tables.materializer(table)(_to_table(result, header))
 end
 
@@ -110,7 +114,17 @@ is appended to `A` along the `append_dim` dimension. The remaining `kwargs` corr
 the usual [`Transform`](@ref) being invoked.
 """
 function apply_append(A::AbstractArray, t; append_dim, kwargs...)::AbstractArray
-    return cat(A, apply(A, t; kwargs...); dims=append_dim)
+    result = _apply_append(cardinality(t), A, t; append_dim=append_dim, kwargs...)
+    return cat(A, result; dims=append_dim)
+end
+
+_apply_append(::Cardinality, A, t; kwargs...) = apply(A, t; kwargs...)
+
+function _apply_append(::ManyToOne, A, t; append_dim, kwargs...)
+    # A was reduced along the append_dim so we must reshape the result setting that dim to 1
+    new_size = collect(size(A))
+    setindex!(new_size, 1, dim(A, append_dim))
+    return reshape(apply(A, t; kwargs...), new_size...)
 end
 
 """
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -37,13 +37,13 @@ instead of being relative to a certain dimension.
 function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
     if dims === Colon()
         if inds === Colon()
-            return _apply(cardinality(t), A, t; dims=:, kwargs...)
+            return _apply(_preformat(cardinality(t), A, :), t; dims=:, kwargs...)
         else
-            return _apply(cardinality(t), A[:][inds], t; dims=:, kwargs...)
+            return _apply(_preformat(cardinality(t), A[:][inds], :), t; dims=:, kwargs...)
         end
     end
 
-    return _apply(cardinality(t), selectdim(A, dims, inds), t; dims=dims, kwargs...)
+    return _apply(_preformat(cardinality(t), selectdim(A, dims, inds), dims), t; dims=dims, kwargs...)
 end
 
 """
@@ -75,7 +75,7 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
 
     # We call hcat to convert any Vector components/results into a Matrix.
     # Passing dims=2 only matters for ManyToOne transforms - otherwise it has no effect.
-    result = hcat(_apply(cardinality(t), hcat(components), t; dims=2, kwargs...))
+    result = hcat(_apply(_preformat(cardinality(t), hcat(components), 2), t; dims=2, kwargs...))
     return Tables.materializer(table)(_to_table(result, header))
 end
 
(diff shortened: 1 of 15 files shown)


src/traits.jl
```diff
@@ -0,0 +1,49 @@
+"""
+    type Cardinality
+
+A trait describing the cardinality of a [`Transform`]
+[`OneToOne`](@ref), [`ManyToOne`](@ref), [`OneToMany`](@ref), and [`ManyToMany`](@ref).
+"""
+
```
glennmoy, line 7:
```suggestion
```

src/traits.jl
```diff
@@ -0,0 +1,49 @@
...
+Transforms that map many inputs to many outputs: `(x_1, x_2, ..., x_m) → (y_1, y_2, ..., y_n)`.
+Examples: Principle Component Analysis (not implemented).
+"""
+struct ManyToMany <: Cardinality end
+
+
+"""
+    cardinality(transform) -> Cardinality
+
+Return the [`Cardinality`](@ref) of the `transform`.
```
glennmoy, line 47:
```suggestion
Returns the [`Cardinality`](@ref) of the `transform`.
```

test/one_hot_encoding.jl
```diff
@@ -174,14 +174,14 @@
 
         df = DataFrame(:a => ["foo", "bar"], :b => ["foo2", "bar2"])
         expected = DataFrame(
-            [[1, 0], [0, 1], [0, 0], [0, 0], [0, 0], [0, 0], [0, 0], [0, 0], [1, 0], [0, 1]],
-            [Symbol.(:Column, x) for x in 1:10],
+            Bool[1 0 0 0 0; 0 1 0 0 0; 0 0 0 1 0; 0 0 0 0 1],
+            [Symbol.(:Column, x) for x in 1:5],
```
glennmoy, line 178: interestingly this use-case changed when I made the refactoring. I'm not too concerned about this because the previous result was inconsistent with the matrix application. This version is now consistent.

Secondly, I'm not sure how much sense to apply OHE  to multiple columns. The output isn't necessarily valid or useful, so I would considering dropping the tests on multi-dimensional array altogether.

bencottier: Could you explain more what was going on before and how it works now? I'm just pressed for time.

src/apply.jl
```diff
@@ -110,7 +109,9 @@ is appended to `A` along the `append_dim` dimension. The remaining `kwargs` corr
 the usual [`Transform`](@ref) being invoked.
 """
 function apply_append(A::AbstractArray, t; append_dim, kwargs...)::AbstractArray
-    return cat(A, apply(A, t; kwargs...); dims=append_dim)
+    result = apply(A, t; kwargs...)
+    result = _postformat(cardinality(t), result, A, append_dim)
```
bencottier, line 113: Just want to check that `append_dim` is definitely right. This came up in the `apply_append` PR but do we have any test cases for `append_dim != dims` yet?

bencottier (commented):
Overall looks good to me. Generalising LC is nice and I'm glad an intermediate `_apply` wasn't needed. There could still be a risk of `_preformat` and `_postformat` becoming bloated, but I think we can cross that bridge if we come to it.

src/test_utils.jl
```diff
@@ -0,0 +1,25 @@
...
+FeatureTransforms.cardinality(::FakeOneToManyTransform) = OneToMany()
+FeatureTransforms._apply(A, ::FakeOneToManyTransform; kwargs...) = hcat(ones(size(A)), ones(size(A)))
+
+struct FakeManyToOneTransform <: Transform end
+FeatureTransforms.cardinality(::FakeManyToOneTransform) = ManyToOne()
+FeatureTransforms._apply(A, ::FakeManyToOneTransform; dims, kwargs...) = ones(size(first(A)))
+
+struct FakeManyToManyTransform <: Transform end
+FeatureTransforms.cardinality(::FakeManyToManyTransform) = ManyToMany()
+FeatureTransforms._apply(A, ::FakeManyToManyTransform; kwargs...) = hcat(ones(size(A)), ones(size(A)))
```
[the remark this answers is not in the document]
glennmoy: yes - FakeXToManyTransforms should output multiple components. It doesn't really matter what the input is, although for ManyToMany the output _can_ be smaller than the input.

src/apply.jl
```diff
@@ -125,3 +126,24 @@ function apply_append(table, t; kwargs...)
...
+# Before applying a ManyToOne Transform we must first slice up the data along the dimension
+# we are reducing over.
+_preformat(::Cardinality, A, d) = A
+_preformat(::ManyToOne, A, d) = eachslice(A; dims=d)
+
+# _postformat formats the data after calling _apply. Needed for apply_append.
+# After applying a ManyToOne Transform, if we want to cat the result we have to reshape it
+# setting the reduced dimension to 1, otherwise cat will error.
+_postformat(::Cardinality, result, A, d) = result
+function _postformat(::ManyToOne, result, A, d)
```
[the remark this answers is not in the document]
glennmoy: sure, I renamed `d`->`append_dim` in the other PR but I'll explain here as well.

Basically, the problem was when we want to append a row to an array, but LinearCombination always returns a column vector. So we have to reshape it into a row to get it to fit.

In general, when we want to `cat` the result with `apply_append` for ManyToOne transforms (which always return a N-1 array) we have to reshape it so that the reduced dimension is always 1 so that it fits.

src/apply.jl
```diff
@@ -75,9 +73,12 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     # Extract a columns iterator that we should be able to use to mutate the data.
     # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
     coltable = Tables.columntable(table)
-    cols = _to_vec(cols)
+    components = reduce(hcat, getproperty(coltable, col) for col in _to_vec(cols))
```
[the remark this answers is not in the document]
glennmoy: the problem is `getproperty(coltable, col) for col in _to_vec(cols)` is a `Vector{Vector}` so calling `Matrix` on this wouldn't work. 

I assume calling `Matrix` is what you meant by there being a function?

test/linear_combination.jl
```diff
@@ -21,44 +21,36 @@
...
-            @test FeatureTransforms.apply(M, lc) == [2, 4]
-            @test lc(M) == [2, 4]
-        end
-
         @testset "dims" begin
             @testset "dims = :" begin
                 M = [1 1; 2 2; 3 5]
                 lc = LinearCombination([1, -1, 1])
-                @test_throws ArgumentError FeatureTransforms.apply(M, lc; dims=:)
+                @test_throws MethodError FeatureTransforms.apply(M, lc; dims=:)
```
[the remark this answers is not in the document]
glennmoy: When removed the explicit [`check for dims=:`](https://github.com/invenia/FeatureTransforms.jl/blob/aa43348f547339ac62302c8e344e0dfba26480a6/src/linear_combination.jl#L26), so the `ArgumentError` is no longer hit.

Now instead it hits `eachslice(A; dims=:)` which throws a `MethodError`.

src/apply.jl
```diff
@@ -75,9 +73,12 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
...
 
-    result = reduce(hcat, [_apply(getproperty(coltable, col), t; kwargs...) for col in cols])
+    # We call hcat to convert any Vector components/results into a Matrix.
+    # Passing dims=2 only matters for ManyToOne transforms - otherwise it has no effect.
+    input = _preformat(cardinality(t), hcat(components), 2)
+    result = hcat(_apply(input, t; dims=2, kwargs...))
```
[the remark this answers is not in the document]
glennmoy: I've left more comments. It's to make sure we can use `eachslice` in `_preformat`, and then pipe the result into a Table. Without `hcat(components)` we can't take a LinearCombination of a single column in a NamedTuple/DataFrame. Not sure how useful that is but we're supporting it.

test/one_hot_encoding.jl
glennmoy: I think what was happening before is (in the Tables `apply`) when we had multiple columns we were calling `_apply` on each component in turn and then stacking the result. This lead to the original version that had many columns, which would have had duplicate names.

Now instead, we collect the components and call `_apply` on the collection, which is what we were doing for the `Array` method all along. So now we get one column per category (as expected) but the shape is now inconsistent.

This is what I mean by it doesn't make sense to call OHE on a Matrix. I think in a future release I'm going to delete these tests because they don't seem worth supporting.

glennmoy:
closed in favour of https://github.com/invenia/FeatureTransforms.jl/pull/80

PR #79, opened 2021-04-16
Define cardinality traits

glennmoy:
First step towards #75 is to define the various traits and add them to the existing transforms

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FeatureTransforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.3.2"
+version = "0.3.3-DEV"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -1,0 +1,48 @@
+"""
+    type Cardinality
+
+A trait describing the cardinality of a [`Transform`]. Available cardinalities are:
+[`OneToOne`](@ref), [`ManyToOne`](@ref), [`OneToMany`](@ref), and [`ManyToMany`](@ref).
+"""
+abstract type Cardinality end
+
+"""
+    OneToOne <: Cardinality
+
+Transforms that map each input to exactly one output: `x → y`.
+Examples: [`Power`](@ref), [`Periodic`](@ref).
+"""
+struct OneToOne <: Cardinality end
+
+"""
+    ManyToOne <: Cardinality
+
+Transforms that map many inputs to one output: `(x_1, x_2, ..., x_n) → y`.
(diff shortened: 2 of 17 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "FeatureTransforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.3.2"
+version = "0.3.3-DEV"
```
glennmoy, line 4: I'm going to follow up with TestUtils before tagging a release

src/traits.jl
```diff
@@ -0,0 +1,48 @@
...
+"""
+struct ManyToMany <: Cardinality end
+
+
+"""
+    cardinality(transform) -> Cardinality
+
+Returns the [`Cardinality`](@ref) of the `transform`.
+"""
+function cardinality end
```
bencottier, line 48: I initially thought defining `cardinality` on instances rather than types seemed redundant in a way. And that `cardinality` for each subtype of `AbstractScaling` seemed redundant. But I realised it's how traits are supposed to work right?

bencottier (approved):
Looks good to me

src/traits.jl
glennmoy: I'm not sure I understand. 
In this instance I could/should have defined the `cardinality` on `AbstractScaling`, which would have covered `MeanStdScaling` and `IdentityScaling`.

bencottier: Yes that's partly what I was thinking. For some reason thought defining `cardinality` on `AbstractScaling` wouldn't work because it was an abstract type, so I didn't comment on that directly. 

The other point I was making is, why should every instance of a given `Transform` subtype have this trait, when the trait is the same between all instances? Conceptually, cardinality belongs more with the type than the instances.

I backed off on that because I figured traits, as a design pattern, apply to instances. But I'm not even sure about that - is it a valid point?

glennmoy: oh right, well they can apply to both. 

You'll see I've changed the code so that cardinality now applies to `AbstractScaling`.
In that case it made more sense to attach it to the type because all concrete subtypes will have the same behaviour.

But I attached it directly to the concrete types of all other transforms because they don't have abstract supertype (besides Transform) to define it for. And if I defined it for `Transform` outright then users won't know they have to define it themselves.

bencottier: Yep that looks fine. Still not sure if I got this across though:

> why should every instance of a given Transform subtype have this trait, when the trait is the same between all instances? Conceptually, cardinality belongs more with the type than the instances.

This means e.g.

```
julia> cardinality(T) where T <: Power = OneToOne()

julia> cardinality(Power) 
OneToOne
```

vs.

```
julia> cardinality(::Power) = OneToOne()

julia> cardinality(Power(2))
OneToOne
```

PR #80, opened 2021-04-16
Refactor apply methods to use Traits

glennmoy:
Second step in #75 

replaces #77 

Note I haven't changed the LinearCombination tests to prevent this being a breaking change. Instead I added a deprecation warning if `dims=:` is passed in.

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -33,21 +33,19 @@ Provide the `inds` keyword to apply the [`Transform`](@ref) to certain indices a
 
 Note: if `dims === :` (all dimensions), then `inds` will be the global indices of the array,
 instead of being relative to a certain dimension.
-
-This method does not guarantee the data type of what is returned. It will try to conserve
-type but the returned type depends on what the original `A` was, and the `dims` and `inds`
-specified.
 """
 function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
+    c = cardinality(t)
     if dims === Colon()
         if inds === Colon()
-            return _apply(A, t; kwargs...)
+            return _apply(_preformat(c, A, :), t; dims=:, kwargs...)
         else
-            return @views _apply(A[:][inds], t; kwargs...)
+            return _apply(_preformat(c, A[:][inds], :), t; dims=:, kwargs...)
         end
     end
 
-    return _apply(selectdim(A, dims, inds), t; kwargs...)
+    input = _preformat(c, selectdim(A, dims, inds), dims)
+    return _apply(input, t; dims=dims, kwargs...)
 end
 
 """
@@ -75,9 +73,14 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     # Extract a columns iterator that we should be able to use to mutate the data.
     # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
     coltable = Tables.columntable(table)
-    cols = _to_vec(cols)
+    # Combine the Vector{Vector} of components into a Matrix
+    components = reduce(hcat, getproperty(coltable, col) for col in _to_vec(cols))
 
-    result = reduce(hcat, [_apply(getproperty(coltable, col), t; kwargs...) for col in cols])
+    # Calling hcat converts any Vector components/results into a Matrix.
+    # First hcat ensures we can use eachslice, second hcat is so we can put result in a Table.
+    # Passing dims=2 only matters for ManyToOne transforms - otherwise it has no effect.
+    input = _preformat(cardinality(t), hcat(components), 2)
+    result = hcat(_apply(input, t; dims=2, kwargs...))
     return Tables.materializer(table)(_to_table(result, header))
 end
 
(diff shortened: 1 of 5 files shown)


test/one_hot_encoding.jl
```diff
@@ -175,14 +175,14 @@
 
         df = DataFrame(:a => ["foo", "bar"], :b => ["foo2", "bar2"])
         expected = DataFrame(
-            [[1, 0], [0, 1], [0, 0], [0, 0], [0, 0], [0, 0], [0, 0], [0, 0], [1, 0], [0, 1]],
```
glennmoy, line 178: These tests changed because what was happening before (in the Tables apply) when we had multiple columns is that we were calling _apply on each component in turn and then stacking the result. This lead to the original version that had many columns, which would have had duplicate names.

Now instead, we collect the components and call _apply on the collection, which is what we were doing for the Array method all along. So now we get one column per category (as expected) but the shape is now inconsistent.

So the previous result was inconsistent with the matrix application but I didn't notice that before. This version is now consistent. Either way I'm not too concerned about this because I'm not sure how much sense it makes to apply OHE to multiple columns. The output isn't necessarily valid or useful, so I would considering dropping the tests on multi-dimensional array altogether.

test/linear_combination.jl
```diff
@@ -161,13 +153,13 @@
...
         @testset "all inds" begin
-            @test FeatureTransforms.apply(A, lc) == [-3, -3]
-            @test lc(A) == [-3, -3]
+            @test FeatureTransforms.apply(A, lc; dims=1) == [-3, -3]
+            @test lc(A; dims=1) == [-3, -3]
         end
 
         @testset "dims" begin
             @testset "dims = :" begin
-                @test_throws ArgumentError FeatureTransforms.apply(A, lc; dims=:)
```
glennmoy, line 170: When we removed the explicit check for `dims=:` the `ArgumentError` is no longer hit.

Now it hits `eachslice(A; dims=:)` which throws a `MethodError` instead .

src/apply.jl
```diff
@@ -75,9 +73,14 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     # Extract a columns iterator that we should be able to use to mutate the data.
     # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
     coltable = Tables.columntable(table)
-    cols = _to_vec(cols)
-
-    result = reduce(hcat, [_apply(getproperty(coltable, col), t; kwargs...) for col in cols])
+    # Combine the Vector{Vector} of components into a Matrix
```
glennmoy, line 76:
```suggestion
    # Combine the Vector{Vector} of components into an Array, which might still be a Vector
```

test/linear_combination.jl
```diff
@@ -9,8 +9,8 @@
         @testset "all inds" begin
             x = [1, 2]
             lc = LinearCombination([1, -1])
-            @test FeatureTransforms.apply(x, lc) == fill(-1)
-            @test lc(x) == fill(-1)
+            @test FeatureTransforms.apply(x, lc; dims=1) == fill(-1)
```
glennmoy, line 12: LinearCombination no longer has a default `dim`, which is actually breaking come to think of it. Unless we add a deprecation somewhere.

glennmoy: I've added a commit reversing the changes to the linear combination tests and instead added a deprecation warning for LinearCombination which we can remove in the next breaking release.

test/linear_combination.jl
```diff
@@ -60,7 +60,7 @@
             @testset "dims = :" begin
                 M = [1 1; 2 2; 3 5]
                 lc = LinearCombination([1, -1, 1])
-                @test_throws ArgumentError FeatureTransforms.apply(M, lc; dims=:)
+                @test_deprecated FeatureTransforms.apply(M, lc; dims=:)
```
glennmoy, line 63: We have removed the default `dims=1` for LinearCombination but added a deprecation warning so it's non-breaking for now.

src/linear_combination.jl
```diff
@@ -1,72 +1,27 @@
...
-"""
-    apply(
-        ::AbstractArray{<:Real, N}, ::LinearCombination; dims=1, inds=:
-    ) -> AbstractArray{<:Real, N-1}
-
-Applies the [`LinearCombination`](@ref) to each of the specified indices in the N-dimensional
-array `A`, reducing along the `dim` provided. The result is an (N-1)-dimensional array.
-
-The default behaviour reduces along the column dimension.
+When applied to an N-dimensional array, `LinearCombination` reduces along the `dim` provided
```
bencottier, line 6: I think we should add that the `dim` shouldn't be `:` (something to that effect). Currently this is catched by the deprecation warning, but we'll still need to handle `dims=:` once that is removed.

src/apply.jl
```diff
@@ -33,21 +33,29 @@ Provide the `inds` keyword to apply the [`Transform`](@ref) to certain indices a
...
 Note: if `dims === :` (all dimensions), then `inds` will be the global indices of the array,
 instead of being relative to a certain dimension.
-
-This method does not guarantee the data type of what is returned. It will try to conserve
-type but the returned type depends on what the original `A` was, and the `dims` and `inds`
-specified.
 """
 function apply(A::AbstractArray, t::Transform; dims=:, inds=:, kwargs...)
+
+    if t isa LinearCombination && dims === Colon()
```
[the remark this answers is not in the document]
bencottier: Saw you added a TODO in #81

src/apply.jl
bencottier: Is it because you can get e.g. `[[1], [2], [3]]` which is a `Vector{Vector}` but becomes a `Vector` as `[1, 2, 3]`?

bencottier (approved):
Looks good to me, I think the `Tables.matrix` solution is better than the previous one - simpler and more readable.

Issue #74, opened 2021-04-07
Better way to check is_transformable?


glennmoy:
The transform API is currently checked by calling `is_transformable` on a type, which just checks for types that are expected to have an `apply` method defined for it.

This is rather brittle, since if users define an `apply` method for their type downstream but forget to extend `is_transformable` for it, and unnecessary.
Rather, `is_transformable(my_type)` should just verify that an `apply` method for `my_type` exists and cut out this middle step.

bencottier:
I thought

```julia
is_transformable(::T) where T = hasmethod(apply, Tuple{T, Transform})
```

But that allows anything for `T` because of the table method. How do we deal with that?

glennmoy:
> I thought
> 
> ```julia
> is_transformable(::T) where T = hasmethod(apply, Tuple{T, Transform})
> ```
> 
> But that allows anything for `T` because of the table method. How do we deal with that?

We can probably work around it, for example `methodswith` can restrict to _not_ checking against the supertypes.
So in this case something like this would work:

```julia
function is_transformable(T::Type) 
    Tables.istable(T) && return true  # cannot check against the method with no type annotation
    return !isempty(methodswith(T, FeatureTransforms.apply; supertypes=true))
end
```

```julia
julia> trans = [[1, 2, 3], [[1 2], [3 4]], (a=[1, 2], b=[3, 4]), DataFrame(:a=>[1, 2], :b=>[3, 4])];

julia> all(is_transformable(t) for t in trans)
true

julia> all(is_transformable(typeof(t)) for t in trans)
true
```

But I'm not sure how effective it is to rely on the method table for these things.

PR #81, opened 2021-04-16
Add test fakes and test utils

glennmoy:
Last part of #75 

After this we can tag a release. In a follow-up PR I will refactor the tests to make use of the fakes.

Note: I also moved `is_transformable` into the TestUtils and added a deprecation.


src/apply.jl
```diff
@@ -83,7 +83,7 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     # Extract a columns iterator that we should be able to use to mutate the data.
     # NOTE: Mutation is not guaranteed for all table types, but it avoid copying the data
     coltable = Tables.columntable(table)
-    # Combine the Vector{Vector} of components into a Matrix
+    # Consolidate the Vector{Vector} components into a single array, which might be a Vector.
```
bencottier, line 86: I commented on this in https://github.com/invenia/FeatureTransforms.jl/pull/80#discussion_r614855184 since it's relevant to my other comments there

src/test_utils.jl
```diff
@@ -0,0 +1,56 @@
...
+export FakeManyToOneTransform, FakeManyToManyTransform
+
+for C in (:OneToOne, :OneToMany, :ManyToOne, :ManyToMany)
+    FT = Symbol(:Fake, C, :Transform)
+    @eval begin
+        """
+            $($FT) <: Transform
+
+        A fake [`$($C)`](@ref) transform for test purposes.
+        Calling `apply` will return an array of ones with the expected size and dimension.
```
bencottier, line 23:
```suggestion
        Calling `apply` will return an array of ones with a size and dimension matching the cardinality of the transform.
```
(note line length)

"expected size" sounds a bit strange because the exact size doesn't matter

docs/src/transform_interface.md
```diff
@@ -2,16 +2,18 @@
...
 The "transform interface” is a mechanism that allows sequences of `Transform`s to be combined (with other steps) into end-to-end feature engineering pipelines.
 
-This is supported by the return of a `Transform`s having the same type as the input.
+This is supported by the return of a `Transform` having the same type as the input.
 This type consistency helps to make `Transform`s _composable_, i.e., the output of one is always a valid input to another, which allows users to "stack" sequences of `Transform`s together with minimal glue code needed to keep it working.
 
 Morever, the end-to-end pipelines themselves should obey the same principle: you should be able to add or remove `Transform`s (or another pipeline) to the output without breaking your code.
 That is, the output should also be a valid "transformable" type: either an `AbstractArray`, a `Table`, or other type for which the user has extended [`FeatureTransforms.apply`](@ref) to support.
 Valid types can be checked by calling `is_transformable`, which is the first part of the transform interface.
+See the [FeatureTransforms.TestUtils](@id test-utils) for these testing utiliies.
```
bencottier, line 11: I don't see what `these testing utilities` is referring to - what about

```suggestion
See the [FeatureTransforms.TestUtils](@id test-utils) for this and other testing utiliies.
```

src/test_utils.jl
```diff
@@ -0,0 +1,56 @@
...
+export FakeOneToOneTransform, FakeOneToManyTransform
+export FakeManyToOneTransform, FakeManyToManyTransform
+
+for C in (:OneToOne, :OneToMany, :ManyToOne, :ManyToMany)
+    FT = Symbol(:Fake, C, :Transform)
+    @eval begin
+        """
+            $($FT) <: Transform
+
+        A fake [`$($C)`](@ref) transform for test purposes.
```
bencottier, line 22: I think we should say more here on why the fake transforms are named according to cardinality, and why that's useful for testing.

src/test_utils.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+
+"""
+    is_transformable(x)
+
+Determine if `x` is both a valid input and output of any [`Transform`](@ref), i.e. that it
+follows the [`transform`](@ref) interface.
+Currently, all subtypes of `Table`s and `AbstractArray`s are transformable.
+"""
+is_transformable(::AbstractArray) = true
```
bencottier, line 53: Just wondering how other packages will implement this syntax-wise.

Does this work

```
using FeatureTransforms.TestUtils: is_transformable

is_transformable(::MyType) = true
```

Or does it have to be this

```
using FeatureTransforms

FeatureTransforms.TestUtils.is_transformable(::MyType) = true
```

bencottier (approved):
Looks good, a few minor suggestions and comments

bencottier:
You'll have to update the doctests: https://github.com/invenia/FeatureTransforms.jl/pull/81/checks?check_run_id=2362057980#step:6:222

src/test_utils.jl
glennmoy: the second option I would say, but we'd like to move away from users defining it themselves #74

PR #83, opened 2021-04-20
Use methodswith to determine is_transformable

### LINKED ISSUE #74: Better way to check is_transformable?

glennmoy:
Closes #74 

Having users overload `is_transformable` is error-prone and may end up defeating the purpose of having the `transform` interface in the first place.

Instead it should rely on duck-typing, that is, once `apply` method has been defined for a type it `is_transformable` by definition.


src/test_utils.jl
```diff
@@ -56,10 +57,13 @@ end
...
 Determine if `x` is both a valid input and output of any [`Transform`](@ref), i.e. that it
-follows the [`transform`](@ref) interface.
-Currently, all subtypes of `Table`s and `AbstractArray`s are transformable.
+has an `apply` method defined and therefore follows the [`transform`](@ref) interface.
 """
-is_transformable(::AbstractArray) = true
-is_transformable(x) = Tables.istable(x)
+function is_transformable(T::Type)
+    Tables.istable(T) && return true  # cannot directly check against the method with no type
+    return !isempty(methodswith(T, FeatureTransforms.apply; supertypes=true))
```
glennmoy, line 64: do we expect types to also extend `apply!` and `apply_append`? Personally I think `apply` is enough.

glennmoy, line 64: as @bencottier also pointed out in #74 we can't use `hasmethod` since it would return `true` for any type due to the method we wrote for Tables.

morris25: I think apply is enough.

morris25: `methodswith` doesn't account for the position of `T` within the `apply` arguments so this function claims that all `Transforms` are also transformable.

morris25 (changes requested):

src/test_utils.jl
glennmoy: hmm...is that weird? I could amend it so that `is_transformable(::Transform) == false`?

morris25: > I could amend it so that is_transformable(::Transform) == false?

That seems more correct.  Otherwise it would seem like you could do something like `FeatureTransforms.apply(Power(2), Power(3))` which is incorrect. I'm not sure if there's a situation in which a `Transform` might be transformable though.

glennmoy: yeah that makes sense - I've changed it now

morris25 (approved):

PR #84, opened 2021-04-20
Refactor tests to test types and transform functionality separately

glennmoy:
Note: I wrote the individual testsets for each data type before going around and removing them from each transform. In some cases this removed all the code so I left behind some unit tests specific to the transform.

In a follow up PR or commit I'd like to refactor the transform tests, but that is out of scope for this PR.


test/types/tables.jl
```diff
@@ -0,0 +1,171 @@
+@testset "$TableType" for TableType in (columntable, DataFrame)
```
glennmoy, line 1: `rowtable` actually doesn't work with `apply!` because we transform to a `columntable` first. So this will have to be revisited

glennmoy:
Not sure why tests are failing considering https://github.com/invenia/FeatureTransforms.jl/pull/93 (which is branched from this) are passing ...

test/types/cube.jl
```diff
@@ -0,0 +1,129 @@
...
+            T = FakeOneToManyTransform()
+            @test FeatureTransforms.apply(x, T) == ones(2, 6, 4)
+            @test FeatureTransforms.apply(x, T; inds=[1, 2]) == ones(2, 2)
+            @test FeatureTransforms.apply(x, T; dims=2, inds=[2, 3]) == ones(2, 4, 4)
+        end
+
+        @testset "ManyToOne" begin
+            T = FakeManyToOneTransform()
+            @test FeatureTransforms.apply(x, T; dims=1) == ones(3, 4)
+            @test FeatureTransforms.apply(x, T; dims=1, inds=[1, 2]) == ones(3, 4)
```
bencottier, line 25: Could test just one index here rather than 2, for something different, because the first dim only has size 2

test/types/xarray.jl
```diff
@@ -0,0 +1,137 @@
...
+
+    x = ArrayType([1 2 3; 4 5 6], foo=["a", "b"], bar=[:x, :y, :z])
+
+    @test is_transformable(x)
+
+    @testset "apply" begin
+
+        @testset "OneToOne" begin
+            T = FakeOneToOneTransform()
+            @test FeatureTransforms.apply(x, T) == ones(2, 3)
```
bencottier, line 11: These transforms don't preserve type?

test/types/xarray.jl
```diff
@@ -0,0 +1,137 @@
...
+        @testset "indexing with dims" begin
+
+            T = FakeOneToOneTransform()
+
+            # apply
+            @test FeatureTransforms.apply(x, T; dims=:foo) == ones(2, 3)
+            @test FeatureTransforms.apply(x, T; dims=:bar) == ones(2, 3)
+
+            # apply to certain dims/inds
+            @test FeatureTransforms.apply(x, T; dims=:bar, inds=Key([:x, :z])) == ones(2, 2)
```
bencottier, line 127: Oh nice, didn't know this worked

test/types/xarray.jl
```diff
@@ -0,0 +1,137 @@
...
+
+            @test isequal(
+                FeatureTransforms.apply_append(x, T; inds=[1, 2], append_dim=2),
+                hcat(x, ones(2, 2))
+            )
+        end
+
+    end
+
+    if ArrayType == KeyedArray
```
bencottier, line 117: I forget, are there issues with AxisArray here? Or are you just not bothering?

bencottier (approved):
LGTM. Nice job. -1000+ lines 🤤

test/types/tables.jl
glennmoy: https://github.com/invenia/FeatureTransforms.jl/pull/93

test/types/xarray.jl
glennmoy: The dims syntax for AxisArrays expects an `Axis` type not a `Symbol` so it's incompatible with our interface unless we call `axisdim(A, Axis{:foo})` or something to extract the dimension index.

Something we could do maybe to support this without depending on AxisArrays is:
```julia
if hasfield(A, axes)
    dims = only(findall(a -> a == dims || typeof(a) <: dims, A.axes))
end
```

but I'd prefer to implement that at a later time if we actually need it (and I'd almost prefer to drop AxisArrays than do that).

test/types/xarray.jl
glennmoy: it does in general, and reminder that 
```
julia> KA = KeyedArray(ones(3, 5), features=[:a, :b, :c], time=1:5);

julia> KA == ones(3, 5)
```

but as usual AxisArrays breaks the mould, and any operation on them doesn't return an AxisArray anyway
```julia
julia> A
2-dimensional AxisArray{Int64,2,...} with axes:
    :foo, ["a", "b"]
    :bar, [:x, :y, :z]
And data, a 2×3 Matrix{Int64}:
 1  2  3
 4  5  6

julia> A .^ 3
2×3 Matrix{Int64}:
  1    8   27
 64  125  216

julia> A .+ 1
2×3 Matrix{Int64}:
 2  3  4
 5  6  7
```

another reason to drop it

test/types/xarray.jl
glennmoy: yeah something I learned trying to help Letif out the other day

test/types/matrix.jl
```diff
@@ -0,0 +1,116 @@
...
+        @testset "OneToMany" begin
+            T = FakeOneToManyTransform()
+            @test FeatureTransforms.apply(x, T) == ones(2, 6)
+            @test FeatureTransforms.apply(x, T; inds=[1, 2]) == ones(2, 2)
+            @test FeatureTransforms.apply(x, T; dims=2, inds=[2, 3]) == ones(2, 4)
+        end
+
+        @testset "ManyToOne" begin
+            T = FakeManyToOneTransform()
+            @test FeatureTransforms.apply(x, T; dims=1) == ones(3)
```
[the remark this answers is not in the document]
glennmoy: nope, `_preformat` expects it not to be `Colon()`
```
julia> T = FakeManyToOneTransform()
FakeManyToOneTransform()

julia> A  = ones(4, 3);

julia> T(A)
ERROR: MethodError: no method matching length(::Colon)

File: test/types/vector.jl [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+
+    @testset "apply!" begin
+        T = FakeOneToOneTransform()
+
+        _x = copy(x)
+        FeatureTransforms.apply!(_x, T)
+        @test _x == ones(3)
+
+        # https://github.com/invenia/FeatureTransforms.jl/issues/68
+        @test_broken FeatureTransforms.apply!(copy(x), T; inds=[1, 2])
```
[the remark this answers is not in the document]
glennmoy: wouldn't be a bad idea to have ready at least

test/types/tables.jl
```diff
@@ -0,0 +1,165 @@
...
+                FeatureTransforms.apply(table, T; header=[:x]),
+                TableType((x=ones(3),))
+            )
+        end
+
+        @testset "ManyToMany" begin
+
+            T = FakeManyToManyTransform()
+
+            expected = TableType((
```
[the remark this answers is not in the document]
glennmoy: removed the `expected` pattern

test/types/tables.jl
```diff
@@ -0,0 +1,165 @@
...
+
+            @test isequal(
+                FeatureTransforms.apply(table, T; header=[:p, :q, :r, :s]),
+                TableType((p=ones(3), q=ones(3), r=ones(3), s=ones(3)))
+            )
+        end
+
+    end
+
+    @testset "apply!" begin
```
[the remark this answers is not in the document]
glennmoy: made a note

glennmoy:
I'm gonna force merge this even though tests aren't passing - I'm not sure what's going on but they are passing over in https://github.com/invenia/FeatureTransforms.jl/pull/93 so they should at least be fixed with those.

PR #85, opened 2021-04-21
Ensure TestFakes can return a KeyedArray

glennmoy:
Needed downstream in https://github.com/invenia/AxisSets.jl/pull/50 where we want to extend the transform interface.

Note this doesn't apply to AxisArrays.

### DIFF
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -11,6 +11,13 @@ using FeatureTransforms.TestUtils
 
         M = reshape(1:9, 3, 3)
         @test FeatureTransforms.apply(M, t) == ones(3, 3)
+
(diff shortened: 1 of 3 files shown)


test/test_utils.jl
```diff
@@ -11,6 +11,13 @@ using FeatureTransforms.TestUtils
 
         M = reshape(1:9, 3, 3)
         @test FeatureTransforms.apply(M, t) == ones(3, 3)
+
+        # TODO: remove when refactoring tests
```
glennmoy, line 15: I've already started this https://github.com/invenia/FeatureTransforms.jl/pull/84

morris25 (approved):

PR #86, opened 2021-04-22
Support Julia >=1.3 versions

glennmoy:
should help with integrating into https://github.com/invenia/AxisSets.jl/pull/50

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - "1.5"  # Invenia Prod version
```
glennmoy, line 18: we still add test specifically for 1.5

rofinn (approved):
LGTM

PR #90, opened 2021-05-11
Use std of 0 for singleton vectors

### LINKED ISSUE #89: Scale singleton vectors to [0]
Currently

```julia
julia> MeanStdScaling([1])
MeanStdScaling(1.0, NaN)
```

due to `Statistics.std` behaviour: `corrected=true` by default, which divides the sum of errors by N-1 instead of N, and N-1 is 0 for singletons. 

It's an edge case, but could arise in feature engineering pipelines where the size of available data varies. I think for the purpose of feature engineering, a singleton vector should be standardized to `[0]` rather than `[NaN]`.

bencottier:
Closes #89 by adding an optional kwarg `corrected` which is passed to the `std` function.

### DIFF
--- a/src/scaling.jl
+++ b/src/scaling.jl
@@ -31,8 +31,8 @@ struct MeanStdScaling <: AbstractScaling
     σ::Real
 
     """
-        MeanStdScaling(A::AbstractArray; dims=:, inds=:) -> MeanStdScaling
-        MeanStdScaling(table, [cols]) -> MeanStdScaling
+        MeanStdScaling(A::AbstractArray; dims=:, inds=:, corrected=true) -> MeanStdScaling
+        MeanStdScaling(table, [cols], corrected=true) -> MeanStdScaling
 
     Construct a [`MeanStdScaling`](@ref) transform from the statistics of the given data.
     By default _all the data_ is considered when computing the mean and standard deviation.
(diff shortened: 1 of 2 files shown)


src/scaling.jl
```diff
@@ -68,7 +68,8 @@ struct MeanStdScaling <: AbstractScaling
     end
 end
 
-compute_stats(x) = (mean(x), std(x))
+# Set std to 0 using corrected=false if x is a singleton
+compute_stats(x) = (mean(x), std(x; corrected=(length(x) != 1))
```
bencottier, line 72:
```suggestion
compute_stats(x) = (mean(x), std(x; corrected=(length(x) != 1)))
```

src/scaling.jl
```diff
@@ -68,7 +68,8 @@ struct MeanStdScaling <: AbstractScaling
     end
 end
 
-compute_stats(x) = (mean(x), std(x))
+# Set std to 0 using corrected=false if x is a singleton
+compute_stats(x) = (mean(x), std(x; corrected=(length(x) != 1)))
```
glennmoy, line 72: maybe it's better to pass this via `MeanStdScaling` as a kwarg? since it's already exposed via `Statistics.std` anyway?

bencottier: I see where you're coming from by leaving it up to the user. I guess it would just be inconvenient in the case that came up for me. I would rather change the test case so that the data doesn't end up being one row, than add `corrected=(length(x) != 1)` to `MeanStdScaling` in the transform pipeline, in the source code.

src/scaling.jl
```diff
@@ -31,8 +31,8 @@ struct MeanStdScaling <: AbstractScaling
     σ::Real
 
     """
-        MeanStdScaling(A::AbstractArray; dims=:, inds=:) -> MeanStdScaling
-        MeanStdScaling(table, [cols]) -> MeanStdScaling
+        MeanStdScaling(A::AbstractArray; dims=:, inds=:, corrected=true) -> MeanStdScaling
```
glennmoy, line 34: Can we just pass generic `kwargs...`? Or does it get confused about what to do with the `dims` kwarg?

PR #93, opened 2021-06-21
Support RowTables

### LINKED ISSUE #64: Tests do not check rowtables
We only check NamedTuples and DataFrames both of which are columntables.

glennmoy:
Closes #64 

Note that we cannot call `apply!` on rowtables since the elements of the NamedTuple are immutable in this instance.

e.g. 
```julia
# can mutate the vector elements
coltable = NamedTuple{(:a, :b)}(a=[1, 2, 3], b=[4, 5, 6])

# cannot mutate the integer elements
rowtable = [
    NamedTuple{(:a, :b)}(a=1, b=4), 
    NamedTuple{(:a, :b)}(a=2, b=5), 
    NamedTuple{(:a, :b)}(a=3, b=6),
]
```

Converting to-from a columntable defeats the purpose as it wouldn't be mutating the underlying data in memory.
So we just don't support it.

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -89,23 +89,33 @@ function apply(table, t::Transform; cols=_get_cols(table), header=nothing, kwarg
     return Tables.materializer(table)(_to_table(result, header))
 end
 
+# Prevents Vector{NamedTuple} from using the AbstractArray method above
+function apply(table::Tables.RowTable, t::Transform; kwargs...)
+    return rowtable(apply(Tables.columntable(table), t; kwargs...))
+end
+
 """
     apply!(table::T, ::Transform; [cols])::T where T
 
 Applies the [`Transform`](@ref) to each of the specified columns in the `table`.
 If no `cols` are specified, then the [`Transform`](@ref) is applied to all columns.
+
+!!! Note
+  `apply!` does not support `RowTable`s since the `NamedTuple` rows are immutable.
 """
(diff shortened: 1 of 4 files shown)


test/types/tables.jl
```diff
@@ -1,5 +1,5 @@
 # TODO: test on rowtable https://github.com/invenia/FeatureTransforms.jl/issues/64
```
glennmoy, line 1:
```suggestion
```

glennmoy:
reminder to bump the project

src/apply.jl
```diff
@@ -153,5 +171,5 @@ _postformat(::Cardinality, result, A, append_dim) = result
 function _postformat(::ManyToOne, result, A, append_dim)
     new_size = collect(size(A))
     setindex!(new_size, 1, dim(A, append_dim))
-    return reshape(result, new_size...)
+    return copy(reshape(result, new_size...))  # return a copy to remove the reshape type
```
mjram0s, line 174: Was the reshape type a problem before? Just wondering why we weren't already returning a copy.

mjram0s (approved):
Just one comment but otherwise everything is pretty straightforward and LGTM

src/apply.jl
glennmoy: Yeah it was a problem because calling `parent` on it later on removed the `ReshapeArray` type and returned the underlying `AxisArray` / `KeyedArray` (which is what we actually wanted to call `parent` on). So calling `copy` just performs that step in advance.

PR #98, opened 2021-07-12
use max rather than maximum to decide if need to use epislon

oxinabox:
This saves on allocations


glennmoy (approved):

glennmoy (approved):
Might want to bump the patch as well (since it's an internal optimisation?)

PR #103, opened 2021-08-25
test dependency on DataFrames update compat

mzgubic:
Nightly CI on Julia 1.3 started failing last night (I don't know what change caused it) but essentially the issue is that on older julia versions the package resolver is not willing to downgrade test-time packages only, see more detail here: https://www.oxinabox.net/2021/02/13/Julia-1.6-what-has-changed-since-1.0.html#resolver-willing-to-downgrade-packages-to-install-new-ones-tiered-resolution

It turns out that we can simply update the DataFrames compat to 1 to fix it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "FeatureTransforms"
 uuid = "8fd68953-04b8-4117-ac19-158bf6de9782"
 authors = ["Invenia Technical Computing Corporation"]
-version = "0.3.9"
+version = "0.3.10"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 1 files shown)


thomasgudjonwright:
Sorry if I am misunderstanding, but if 0.22 is problematic should it not be removed from the compat?

mzgubic:
It's sort of complicated, there are a bunch of things to consider.

We only specify compat for the things that would actually break in this package, and since this package (at test time) uses only the basic DataFrames interface, it is compatible with both 0.22 and 1. We don't care if our package depends on a package that breaks in 0.22 - it is that package's job to figure out it's own compat bounds.

What actually happened here though was not that there was anything broken in 0.22. It's just that in old julia versions the package manager was not smart enough to deal with test-only dependencies. It would first build the main environment, and when adding the extra test-only dependencies not willing to downgrade one of the main dependencies.

thomasgudjonwright (approved):
Cool, that makes sense!

PR #104, opened 2021-12-03
Log Transforms

AlexRobson:
Adds two forms of log transforms. 

Either:

* LogTransform: sign(x) * log(|x| + 1) 
* InverseHyperbolicSine: log(x + √(x² + 1)).

### DIFF
--- a/src/FeatureTransforms.jl
+++ b/src/FeatureTransforms.jl
@@ -8,7 +8,7 @@ using Tables
 export Transform, transform, transform!
 export HoD, LinearCombination, OneHotEncoding, Periodic, Power
 export AbstractScaling, IdentityScaling
-export MeanStdScaling, LogTransform
+export MeanStdScaling, LogTransform, InverseHyperbolicSine
 
 include("utils.jl")
 include("traits.jl")
--- a/src/FeatureTransforms.jl
+++ b/src/FeatureTransforms.jl
@@ -7,8 +7,8 @@ using Tables
 
 export Transform, transform, transform!
 export HoD, LinearCombination, OneHotEncoding, Periodic, Power
-export AbstractScaling, IdentityScaling
-export MeanStdScaling, LogTransform, InverseHyperbolicSine
+export AbstractScaling, IdentityScaling, MeanStdScaling
+export LogTransform, InverseHyperbolicSine
 
 include("utils.jl")
 include("traits.jl")
--- a/src/FeatureTransforms.jl
+++ b/src/FeatureTransforms.jl
@@ -7,7 +7,8 @@ using Tables
 
 export Transform, transform, transform!
 export HoD, LinearCombination, OneHotEncoding, Periodic, Power
-export AbstractScaling, IdentityScaling, MeanStdScaling
+export AbstractScaling, IdentityScaling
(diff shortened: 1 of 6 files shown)


src/log.jl
```diff
@@ -0,0 +1,33 @@
...
+    LogTransform <: Transform
+
+Logarithmically transform the data through: sign(x) * log(|x| + 1).
+
+This allows positive domains elements to be transformed. 
+"""
+struct LogTransform <: Transform end
+cardinality(::LogTransform) = OneToOne()
+
+_logtransform(x) = sign(x) * log(abs(x) + 1)
```
mzgubic, line 11:
```suggestion
_logtransform(x) = sign(x) * log(abs(x) + one(x))
```

src/log.jl
```diff
@@ -0,0 +1,33 @@
...
+_invlogtransform(x) = sign(x) * (exp(sign(x) * x) - 1)
```
mzgubic, line 12:
```suggestion
_invlogtransform(x) = sign(x) * (exp(sign(x) * x) - one(x))
```

src/log.jl
```diff
@@ -0,0 +1,33 @@
+"""
+    LogTransform <: Transform
+
+Logarithmically transform the data through: sign(x) * log(|x| + 1).
+
+This allows positive domains elements to be transformed.
```
mzgubic, line 6:
```suggestion
This allows transformations of all real numbers, not just positive ones.
```

test/log.jl
```diff
@@ -0,0 +1,46 @@
...
+            @test cardinality(transform) == OneToOne()
+            @test transform isa Transform
+        end
+
+        @testset "Apply" for (x, y) in ((V1, logV1), (V2, logV2), (M, logM))
+            f = LogTransform()
+            transformed = FeatureTransforms.apply(x, f)
+            @test transformed ≈ y atol=1e-5
+            @test FeatureTransforms.apply(transformed, f; inverse=true) ≈ x atol=1e-5
+        end
```
mzgubic, line 24: can we add tests for symmetry and going through zero?

src/FeatureTransforms.jl
```diff
@@ -7,7 +7,8 @@ using Tables
 
 export Transform, transform, transform!
 export HoD, LinearCombination, OneHotEncoding, Periodic, Power
-export AbstractScaling, IdentityScaling, MeanStdScaling
+export AbstractScaling, IdentityScaling
+export MeanStdScaling, LogTransform, InverseHyperbolicSine
```
mzgubic, line 11:
```suggestion
export AbstractScaling, IdentityScaling, MeanStdScaling
export LogTransform, InverseHyperbolicSine
```
so that linear scalings are together

mzgubic (approved):
Nice, thanks!

test/log.jl
AlexRobson: what do you mean by symmetric here? 

For going through zero i assume you mean testing something like: V1 = [1; 2; 3; 4; 0; 6]

mzgubic: symmetry: explicitly checking that`log_tsf(x) == log_tsf(-x)`
zero: yes

Issue #106, opened 2022-02-18
Separate transform constructors and `fit` method


mzgubic:
# Current state

Some transformations (e.g. `MeanStdScaling`) need data to determine some of the parameters. At present, this is done by passing the data to the constructor.

As a result, it isn't generically possible to exchange the transformation from `MeanStdScaling` to e.g. `LogTransform` in some data preprocessing step, since they have a different API. An example:
```julia
julia> function pipeline(transform_type, data)
           tsf = transform_type(data)
           return FeatureTransforms.apply(data, tsf)
       end
```
works with
```julia
julia> pipeline(MeanStdScaling, rand(10))
```
but not with
```julia
julia> pipeline(LogTransform, rand(10))
ERROR: MethodError: no method matching LogTransform(::Vector{Float64})
Closest candidates are:
  LogTransform() at ~/JuliaEnvs/FeatureTransforms.jl/src/log.jl:8
Stacktrace:
 [1] pipeline(transform_type::Type{LogTransform}, data::Vector{Float64})
   @ Main ./REPL[38]:2
 [2] top-level scope
   @ REPL[40]:1
```
In this simple case we can avoid the issue by instantiating the transforms outside of the pipeline. However, that means that we _have to have access to data when constructing the pipeline_, which feels wrong and unmodular. Even worse, if `MeanStdScaling` is somewhere in the middle of long list of operations, we have to apply the preceding transformations first to instantiate it _all while we construct the pipeline_.

# A possible alternative

We could implement a `StatsBase.fit!(transform, data)` method which handles the fitting part. The fallback method would do nothing, but for `MeanStdScaling` the fitting work currently done in the constructor would be moved to the `fit!` method.

I would also suggest adding a `fitted` boolean to `MeanStdScaling` so that we can error if the `transform` method is called before `fit`.

Since this is breaking we can also take the opportunity to rename `MeanStdScaling` to whatever comes out of #87.

Related to https://github.com/invenia/FeatureTransforms.jl/issues/59

Issue #87, opened 2021-04-26
MeanStdScaling is a bad name


glennmoy:
As noted by @eperim 

>"Standardised" data is a pretty common nomenclature, while "MeanStdScaling" sounds like it could be anything (and the data is not even scaled by the mean, it is shifted by it, only the std is used for scaling).

bencottier:
I think it's fine! "Standardised" sounds to me like it could be many things 😅 With `MeanStdScaling` I know immediately what it means among all the common ways of scaling data (and I thought so before we wrote this Transform).

I do think "Standardised" is better than "Normalised" (which I considered originally) because while that has "Normal" in the name, it can refer to scaling by any kind of norm.

eperim:
I disagree. "Standardisation" is quite common nomenclature. There's even a [special tag for it in Stats Exchange](https://stats.stackexchange.com/questions/tagged/standardization), and you can find the term frequently in [glossaries](https://statisticsbyjim.com/glossary/standardization/), [multiple](https://kharshit.github.io/blog/2018/03/23/scaling-vs-normalization) [blog posts](https://www.statisticshowto.com/standardized-values-examples/), [teaching materials](https://amsi.org.au/ESA_Senior_Years/SeniorTopic4/4h/4h_2content_8.html), and in other packages (e.g. [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html)). These are all from the first page of a Google search.

> With MeanStdScaling I know immediately what it means among all the common ways of scaling data in ML

Naturally you do, because you chose the name, but is it common nomenclature that you expect people in general to be familiar with?

bencottier:
> "Standardisation" is quite common nomenclature.

Thanks for gathering those links, that makes me favour "Standardisation" more.

> Naturally you do, because you chose the name, but is it common nomenclature that you expect people in general to be familiar with?

I copied the name from FeatureEngineering, but yes I had the chance to rename in the original PR and decided to keep it. I think more in terms of "how to identify what this literally does with a relatively short name?" than the nomenclature. You might even disagree that "MeanStdScaling" is best for that purpose, but that's my reasoning.

I commented to show there's a difference of opinion, but I don't feel strongly about this. Nomenclature is important and it's very reasonable to rename it.

molet:
I would also vote for "Standardization" instead of "MeanStdScaling".
Also, would it be possible to add "Normalization" as well to the package?
[...]

Issue #59, opened 2021-03-21
No-arg `MeanStdScaling` constructor


rofinn:
In many cases, we're gonna call `apply` with the same array data that was used to construct the `MeanstdScaling` transform, so why not provide a no-arg constructor and computing the `mean` and `std` on the fly in the `apply` call? I guess that would make the fields `Union{Real, Nothing}`.

glennmoy:
@bencottier might also remember, but I think one reason we favoured immutable struct was to prevent the user from recycling the same `scaling` and potentially getting confused.

But I guess we could allow mutability iff the fields are `nothing`.

bencottier:
> @bencottier might also remember, but I think one reason we favoured immutable struct was to prevent the user from recycling the same scaling and potentially getting confused.

The other reason was saving code. We'd need to overload `apply`. This reason feels weaker than at the time of the original PR. In general I now feel more ok with overloading `apply` to add convenience. Also I think there'd be less overloading necessary after `apply` became the default (#25 ).

AlexRobson:
A no-arg `MeanStdScaling` transform makes sense. We were just looking at swapping around transforms but at present, this means that we would need to handle the cases differently. e.g. comparing MeanStdScaling and LogTransform (recently added, but it applies to most (all?) other transforms that are not `<:AbstractScaling`. 

```
function pipeline(...)
   ...
   transform = MeanStdScaling(x)
   ...
end
```

```
function pipeline(...)
   ...
   transform = LogTransform(x) # Would error
   ...
end
```

As @mzgubic points out, this requires separating out the constructor and the fitting

PR #107, opened 2022-02-28
`StandardScaling` with separate constructor and fit methods to replace `MeanStdScaling`

mzgubic:
Closes https://github.com/invenia/FeatureTransforms.jl/issues/106, closes https://github.com/invenia/FeatureTransforms.jl/issues/59, closes https://github.com/invenia/FeatureTransforms.jl/issues/87

This is made non-breaking by deprecating the `MeanStdScaling`.

To do:
- [x] mention `fit!`  in the docs

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -7,6 +7,8 @@ First we load some hourly weather data:
 ```jldoctest example
 julia> using DataFrames, Dates, FeatureTransforms
 
+julia> using FeatureTransforms: fit!
+
 julia> df = DataFrame(
             :time => DateTime(2018, 9, 10):Hour(1):DateTime(2018, 9, 10, 23),
             :temperature => [10.6, 9.5, 8.9, 8.9, 8.4, 8.4, 7.7, 8.9, 11.7, 13.9, 16.2, 17.7, 18.9, 20.0, 21.2, 21.7, 21.7, 21.2, 20.0, 18.4, 16.7, 15.0, 13.9, 12.7],
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -5,7 +5,7 @@ In the following example, we will imagine we are training a model to predict the
 First we load some hourly weather data:
 
 ```jldoctest example
-julia> using DataFrames, Dates, FeatureTransforms
+julia> using DataFrames, Dates, FeatureTransforms, StatsBase
 
 julia> df = DataFrame(
             :time => DateTime(2018, 9, 10):Hour(1):DateTime(2018, 9, 10, 23),
@@ -77,13 +77,17 @@ julia> output_cols = [:temperature, :humidity];
 ```
 
 For many models it is helpful to normalize the training data.
-We can use `MeanStdScaling` for that purpose.
+We can use `StandardScaling` for that purpose.
 Note that we are mutating the data frame in-place using `apply!` one column at a time.
 
 ```jldoctest example
-julia> temp_scaling = MeanStdScaling(train_df; cols=:temperature);
+julia> temp_scaling = StandardScaling();
 
-julia> hum_scaling = MeanStdScaling(train_df; cols=:humidity);
+julia> fit!(temp_scaling, train_df; cols=:temperature);
+
+julia> hum_scaling = StandardScaling();
+
+julia> fit!(hum_scaling, train_df; cols=:humidity);
 
 julia> FeatureTransforms.apply!(train_df, temp_scaling; cols=:temperature);
 
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -5,7 +5,7 @@ In the following example, we will imagine we are training a model to predict the
 First we load some hourly weather data:
 
 ```jldoctest example
-julia> using DataFrames, Dates, FeatureTransforms, StatsBase
+julia> using DataFrames, Dates, FeatureTransforms
 
 julia> df = DataFrame(
             :time => DateTime(2018, 9, 10):Hour(1):DateTime(2018, 9, 10, 23),
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -76,7 +76,7 @@ julia> test_df = feature_df[end-1:end, :];
 julia> output_cols = [:temperature, :humidity];
 ```
 
-For many models it is helpful to normalize the training data.
+For many models it is helpful to standardise the training data.
 We can use `StandardScaling` for that purpose.
 Note that we are mutating the data frame in-place using `apply!` one column at a time.
 
@@ -115,7 +115,7 @@ julia> FeatureTransforms.apply!(train_df, hum_scaling; cols=:humidity)
                                                         7 rows omitted
 ```
 
-We can use the same `scaling` transform to normalize the test data:
+We can use the same `scaling` transform to standardise the test data:
 
 ```jldoctest example
 julia> FeatureTransforms.apply!(test_df, temp_scaling; cols=:temperature);
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -8,7 +8,6 @@ DocTestSetup = quote
     using DataFrames
     using Dates
     using FeatureTransforms
-    using StatsBase
 end
 ```
 
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -8,6 +8,7 @@ DocTestSetup = quote
     using DataFrames
     using Dates
     using FeatureTransforms
+    using StatsBase
 end
 ```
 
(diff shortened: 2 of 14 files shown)


src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+
+!!! note
+    `fit!(scaling, data)` needs to be called before the transform can be `apply`ed.
+    By default _all the data_ is considered when `fit!`ing the mean and standard deviation.
 """
-struct MeanStdScaling <: AbstractScaling
+mutable struct StandardScaling <: AbstractScaling
     μ::Real
     σ::Real
+    fitted::Bool
```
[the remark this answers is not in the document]
AlexRobson: The mutability is unfortunate - i think one of the issues comments that this is best avoided. As an additional alternative, would a separate type work, and introduce a constructor type? I thnk that i've seen this pattern elsewhere. Then, no `fitted`, no `mutability`. Possibly no `StatsBase`. 

e.g. something like:

```
struct StandardScalingFitted <: AbstractScaling
    μ::Real
    σ::Real
end

struct StandardScaling end
```

I haven't stepped through this though, there may be good reasons not to do this.

docs/src/examples.md
```diff
@@ -77,13 +77,17 @@ julia> output_cols = [:temperature, :humidity];
 ```
 
 For many models it is helpful to normalize the training data.
```
AlexRobson, line 79: normalize and standardize are being conflated here

src/scaling.jl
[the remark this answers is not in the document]
mzgubic: No you are right, this is an important point, and I did consider it and discussed it with @glennmoy, but forgot to raise it in the self-review. 

Points in favour of immutability (from reading [this](https://docs.julialang.org/en/v1/manual/types/#Mutable-Composite-Types))
- It's more efficient as it doesn't need to be allocated on the heap
- It can be packed into arrays more efficiently

Points in favour of mutability:
- Having an extra type sucks
    - there is more code to maintain
    - makes things more convoluted than they need to be and harder to reason about
- I think in this simple case the compiler could be smart enough to avoid heap allocations. From above link: "Mutable values, on the other hand are heap-allocated and passed to functions as pointers to heap-allocated values except in cases where the compiler is sure that there's no way to tell that this is not what is happening."

Somewhat separate but related, I know we do the `Fit` object thing in our Models interface, but I regret we made that choice.

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+
+!!! note
+    If you want the `StandardScaling` to transform your data consistently you should use
+    the same `inds`, `dims`, or `cols` keywords when calling `apply`. Otherwise, `apply`
+    might rescale the wrong data or throw an error.
+"""
+function StatsBase.fit!(ss::StandardScaling, args...; kwargs...)
+    ss.fitted === true && @warn("StandardScaling is being refit, Y?")
+    μ, σ = _fit(ss, args...; kwargs...)
+    ss.μ, ss.σ, ss.fitted = μ, σ, true
```
AlexRobson, line 63: Drop the mutability and do:

```
StandardScaling(μ, σ, true) 
```

?

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+# `Table` keyword arguments
+* `cols`: the columns to use in computing the statistics. Default uses all columns.
+
+!!! note
+    If you want the `StandardScaling` to transform your data consistently you should use
+    the same `inds`, `dims`, or `cols` keywords when calling `apply`. Otherwise, `apply`
+    might rescale the wrong data or throw an error.
+"""
+function StatsBase.fit!(ss::StandardScaling, args...; kwargs...)
+    ss.fitted === true && @warn("StandardScaling is being refit, Y?")
```
AlexRobson, line 61: "Y?" ?

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+    x -> (x - μ) / σ
+
+where μ and σ are the mean and standard deviation of the training data.
+
+!!! note
+    `fit!(scaling, data)` needs to be called before the transform can be `apply`ed.
+    By default _all the data_ is considered when `fit!`ing the mean and standard deviation.
 """
-struct MeanStdScaling <: AbstractScaling
+mutable struct StandardScaling <: AbstractScaling
```
AlexRobson, line 32:
```suggestion
struct StandardScaling <: AbstractScaling
```

src/scaling.jl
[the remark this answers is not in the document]
AlexRobson: Ah, `fit` can just re-create the type but with fitted = true, i.e. within `fit!` use `return StandardScaling(μ, σ, true)`

This woudl lead to using `fit` instead of `fit!`.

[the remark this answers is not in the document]
AlexRobson: Further comment, perhaps this could also be initialised at (0, 1, false) (to drop the type union),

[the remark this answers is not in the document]
mzgubic: Good idea. It might cause issues down the line though, e.g. if we want to use a transform as a part of an immutable model, i.e. the following would not be possible:
```julia
julia> mutable struct Scaling
           a::Float64
       end

julia> struct Model
           s::Scaling
       end

julia> m = Model(Scaling(1.0))
Model(Scaling(1.0))

julia> m.s.a=2.0
2.0

julia> m
Model(Scaling(2.0))
```
I'll try and think of some performance benchmarks to see whether mutability makes a practical difference.

`0.0, 1.0, false` is indeed better.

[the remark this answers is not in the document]
glennmoy: > The alternative is to have
> 
> ```
>      μ::Union{Real, Nothing}
>      σ::Union{Real, Nothing}
> ```

I think this is perfectly reasonable IMO. 
Why add another field when you can encode the state of the transform by the type of its fields directly? 

Especially when you can just trick the code by manually specifying `fitted=true` without actually having fitted it to anything.
And especiallllly initiating it with `(0, 1, false)` when its supposed to be stateless to begin with.

src/fit.jl
```diff
@@ -0,0 +1 @@
+StatsBase.fit!(t::Transform, args...; kwargs...) = return t
```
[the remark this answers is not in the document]
glennmoy: this also isn't tested from what I can tell

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+* `inds=:`: the indices to use in computing the statistics. Default uses all indices.
+
+# `Table` keyword arguments
+* `cols`: the columns to use in computing the statistics. Default uses all columns.
+
+!!! note
+    If you want the `StandardScaling` to transform your data consistently you should use
+    the same `inds`, `dims`, or `cols` keywords when calling `apply`. Otherwise, `apply`
+    might rescale the wrong data or throw an error.
+"""
```
glennmoy, line 59: I think I'd attach a generic docstring to the generic `fit!` method and then add any specifics about `StandardScaling` to the constructor leaving this docstring empty.

Then we won't have N `fit!` docstrings when someone tries to look it up `>julia ? fit!`

src/scaling.jl
glennmoy: also should we error if the transform is trying to be refit?

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+function StatsBase.fit!(ss::StandardScaling, args...; kwargs...)
+    ss.fitted === true && @warn("StandardScaling is being refit, Y?")
+    μ, σ = _fit(ss, args...; kwargs...)
+    ss.μ, ss.σ, ss.fitted = μ, σ, true
+    return ss
 end
 
-compute_stats(x) = (mean(x), std(x))
```
glennmoy, line 71: don't you need this?

src/scaling.jl
```diff
@@ -17,64 +17,73 @@ IdentityScaling(args...) = IdentityScaling()
...
+function _fit(::StandardScaling, data::AbstractArray; dims=:, inds=:)
+    return if dims isa Colon
+        compute_stats(data)
+    else
+        compute_stats(selectdim(data, dims, inds))
+    end
```
glennmoy, line 72:
```suggestion
    dims isa Colon && return compute_stats(data)
    compute_stats(selectdim(data, dims, inds))
```

glennmoy:
In general I'm in favour of these changes - I just don't think we should extend `StatsBase` without good reason, and I'd prefer go with a `Union` type in the constructor.

Oh and we should also document `fit!` as part of the API and expect stateful transforms to have a method for it as well as `apply`.

src/scaling.jl
mzgubic: >"Y?" ? Is this a shortcut for requesting user input?

sorry not sure why I wrote that, it is indeed short for "why".

We could error just to be safe (made this change), though there isn't anything wrong in principle about reusing the same object to fit new data.

mzgubic:
Thank you both for reviewing, I think it's ready to go now. We went with the mutable approach so as not to inflate the API with an extra `Fit` class.

glennmoy (approved):

docs/src/transforms.md
```diff
@@ -20,6 +22,15 @@ For example, the following defines a squaring operation (i.e. raise to the power
 julia> p = Power(2);
 ```
 
+A stateful transform, such as a [`StandardScaling`](@ref) should also be fit to the data before it is applied:
+```julia-repl
+julia> s = StandardScaling();
+
+julia> x = rand(5);
+
+julia> FeatureTransforms.fit(s, x);
```
mzgubic, line 31:
```suggestion
julia> FeatureTransforms.fit!(s, x);
```

PR #108, opened 2022-05-17
RFC: `Composite` transform

### LINKED ISSUE #105: Composite Transforms
Should we add a `CompositeTransform`?

e.g. if we want a `MeanStdScaling` after a `LogTransform` we could write `MeanStdScaling ∘ LogTransform`.

I don't think the API can support this in its present form, since it does not distinguish between constructing the transforms and fitting them.

mzgubic:
Closes #105 

In the present form, it is limited to `OneToOne()` component transforms, which have the same keyword arguments when calling `apply`.

In order to generalise `Composite` transforms the keyword arguments should become a part of the constructor rather than being passed in the `apply` call.

Todo:
- [ ] add this stuff to the docs

### DIFF
--- a/src/composite.jl
+++ b/src/composite.jl
@@ -20,7 +20,7 @@ true
 struct Composite <: Transform
     transforms::Tuple{Vararg{Transform}}
 
-    function Composite(transforms::Vector{<:Transform})
+    function Composite(transforms::Tuple{Vararg{Transform}})
         all(==(OneToOne()), map(cardinality, transforms)) && return new(transforms)
         throw(ArgumentError("Only OneToOne() transforms are supported."))
     end


src/composite.jl
```diff
@@ -0,0 +1,54 @@
...
+julia> id = IdentityScaling();
+
+julia> power = Power(2.0);
+
+julia> id ∘ power == Composite([power, id])
+true
+```
+"""
+struct Composite <: Transform
+    transforms::Vector{<:Transform}
```
glennmoy, line 21:
```suggestion
    transforms::Tuple{Vararg{Transform}}
```

more memory efficient, and allows for an external constructor like
```julia
Composite(transforms::Transform...) = Composite(transforms)
```

src/traits.jl
```diff
@@ -46,3 +46,13 @@ struct ManyToMany <: Cardinality end
...
+
+Base.:(∘)(::OneToOne, ::OneToOne) = OneToOne()
+Base.:(∘)(::OneToMany, ::OneToOne) = OneToMany()
+Base.:(∘)(::ManyToOne, ::OneToMany) = OneToOne()
+Base.:(∘)(::ManyToMany, ::OneToMany) = OneToMany()
+Base.:(∘)(::OneToOne, ::ManyToOne) = ManyToOne()
+Base.:(∘)(::OneToMany, ::ManyToOne) = ManyToMany()
+Base.:(∘)(::ManyToOne, ::ManyToMany) = ManyToOne()
+Base.:(∘)(::ManyToMany, ::ManyToMany) = ManyToMany()
+Base.:(∘)(c2::Cardinality, c1::Cardinality) = throw(ArgumentError("$c2 ∘ $c1 is undefined."))
```
glennmoy, line 58:
```suggestion
function Base.:(∘)(c2::Cardinality, c1::Cardinality)
    return throw(ArgumentError("Cannot compose cardinalities: $c2 ∘ $c1."))
end
```

src/composite.jl
```diff
@@ -0,0 +1,54 @@
+"""
+   Composite <: Transform
+
+A `Composite` transform is a composition of `Transform`s, currently limited to `OneToOne()`
+cardinality. It can be fit and applied in a single step.
+
+The transforms in `Composite([t1, t2, t3])` are applied in `t1`, `t2`, `t3` order, where
+the output of `t1` is the input to `t2` etc. When using `∘` to create transforms, the order
+is `t3 ∘ t2 ∘ t1`, as in function composition.
```
glennmoy, line 9: hmm...that might get confusing...

src/traits.jl
```diff
@@ -46,3 +46,13 @@ struct ManyToMany <: Cardinality end
 Returns the [`Cardinality`](@ref) of the `transform`.
 """
 function cardinality end
+
+Base.:(∘)(::OneToOne, ::OneToOne) = OneToOne()
```
glennmoy, line 50: I'm not sure if `∘` is the right syntax here... it's used to compose functions whereas here we're sort of "reducing" over the cardinalities.
I guess it makes sense in the context of composing the equivalent transforms?

If we do go for it then `∘` becomes part of the `Cardinality` API and should be documented.

test/composite.jl
```diff
@@ -0,0 +1,45 @@
...
+        @test id ∘ id == Composite([id, id])
+        @test id ∘ id ∘ power == Composite([power, id, id])
+        @test power ∘ id ∘ power == Composite([power, id, power])
+
+        @test power ∘ (id ∘ logt) == Composite([logt, id, power])
+        @test (power ∘ id) ∘ logt == Composite([logt, id, power])
+        @test (power ∘ id) ∘ (logt ∘ id) == Composite([id, logt, id, power])
+
+        @test_throws ArgumentError id ∘ LinearCombination([1, 2, 3])
+        @test_throws ArgumentError OneHotEncoding([1, 2]) ∘ id
```
glennmoy, line 15: alternatively we could use `test_skip` until these are implemented?

src/composite.jl
mzgubic: Yeah I had it the other way around first but the issue then is that the `c.transforms[1]` is the last transform that is applied, which is even more confusing I think.

The only totally non-confusing way is to get rid of the ` ∘ ` syntactic sugar, which is what switches the order.

src/traits.jl
mzgubic: The alternative is to have an internal `_compose` function which would do this instead? I don't think users need to use this at all.

test/composite.jl
mzgubic: I wanted to make sure there is an actual error, so possibly @test_broken is the right thing

src/composite.jl
glennmoy: >The only totally non-confusing way is to get rid of the ∘ syntactic sugar, which is what switches the order.

I wouldn't be against this tbh, but it might be handy for building pipelines of transforms like
```julia
tc = Composite(t1, t2)
...
...
tc = t3 ∘ tc
...
tc = t4 ∘ tc
```

so I can see value in it being used that way

src/traits.jl
glennmoy: yeah making it internal would indicate it shouldn't be used publically

PR #110, opened 2022-08-03
Drop support for Julia 1.5 and older

rofinn:
This seemed like the easier options since Documenter dropped these Julia version a couple weeks ago and is now causing compat issues. 

https://github.com/invenia/FeatureTransforms.jl/actions/runs/2786559009

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - "1.3"
+          - "1.6"
           - "1"    # Latest Release
         os:
           - ubuntu-latest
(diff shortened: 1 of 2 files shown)


rofinn:
Looks like coverage reporting might be different across Julia releases? Also, I think the Julia 1.5 jobs is stuck cause it's required on `main`? I believe that should go away once merged?

Wynand:
This seems to work when the Julia 1.5 test isn't included: https://github.com/invenia/FeatureTransforms.jl/actions/runs/2791698474

And coverage seems to only be an issue because there are 5 less lines of code: https://github.com/invenia/FeatureTransforms.jl/pull/110/checks?check_run_id=7658996965

mzgubic (approved):
I'm fine with dropping support for 1.5.

glennmoy (approved):

AlexRobson:
Small question - Is this actually a non-breaking release? 

I had done something similar for 0.3 for NamedDims and someone commented that it's a minor release. 

https://github.com/invenia/NamedDims.jl/pull/202#discussion_r937572855

(I've got to here because I'm trying to see which packages need NamedDims 1 and I also encountered this package due to a resolution issue from one of our internal packages)

mzgubic:
>Small question - Is this actually a non-breaking release?

Yes, indeed it is, sorry if it caused trouble. I imagine it can be resolved by bumping compat in a few places?

PR #113, opened 2022-10-31
Test application of `IdentityScaling()` does nothing

mjp98:
Perhaps this could be tested for each type of input in `tests/types`?


mzgubic (approved):
Yeah we could add one or two other cases, but there is no need to test exhaustively.

glennmoy (approved):

